PR #11, opened 2019-03-18
Change GPIO4 to BL0937 CF (type 134).

ZericE:
Based on a hardware tear down, the power monitoring chip is really a BL0937, not a HLW8012. The only thing this changes is GPIO4. Making this change will cause the default calibration to be much closer and better centered in the calibration range.


blakadder:
Thank you for your help

PR #21, opened 2019-04-16
Update shelly25 template

curzon01:
- build-in led must be inverted
- GPIO16 must be set to ADE7953 IRQ (156)


blakadder:
Thanks

PR #24, opened 2019-05-01
Interim Power Reading Fix

meingraham:
This template entry will require updating after Theo releases the full support of the Shelly 1PM. For now, a working code fix alternative is provided.


blakadder:
cool!

PR #30, opened 2019-05-17
Update NAME

spike411:
From 'NeoCool Cam 16' to 'Neo Coolcam 16' to make it more consistent with the original 'Neo Coolcam' template.


blakadder:
Thanks

PR #35, opened 2019-05-23
Update a mistake in gpio type

ProfEngr:
After further testing w/ higher loads I discovered gpio12 should have been 131 instead of 130. I tested w/ a 60W incandescent lamp and a 600W toaster.


blakadder:
Thank you

PR #36, opened 2019-05-26
Update shelly_25

Christian308:
Shelly 2.5 does not have selectable input (only Shelly 1 does).


blakadder:
Thanks for contributing

PR #38, opened 2019-06-02
Create Jinvoo Smart WiFi Valve SM-AW713

adriant0ie:
I stumbled upon JinvooSmart SM-PW713 template, but using that on the JinvooSmart SM-AW713, would keep reseting the device back to basic WiFi hotspot after a few switch ON/OFF.
I tried a few other available templates and with the "ESP switch" I could control this valve (using the 4th button), so I modified the JinvooSmart SM-PW713 to make use of the GPIOs from "ESP switch" one.
So, this template keeps SM-AW713 stable and fully functional (including the toggle button on the valve).

Thanks
Adrian


blakadder:
Thank you for your contribution

PR #47, opened 2019-06-15
Update neocoolcam_16a

VivantSenior:
Added new information


blakadder:
this is not correct, you're putting a description inside the header

PR #57, opened 2019-07-08
Sonoff Basic R3

meingraham:


blakadder:
R3 should be a separate device since it looks completely different

meingraham:
Do you mean externally different appearance? They are all still "the" Sonoff Basic to users. There are still some R1/R2 in the wild, but Itead has discontinued those. Should this become "the" Basic and we relegate the R1/R2 to "legacy" status?

Should the same philosophy (separate template entries) be applied across all three revisions? Internally, R1/R2 were quite different. The R1/R3 base PCB resemble each other with the use of solder rails. The change being the piggyback PCB on the R3.

Mike

blakadder:
yes, they're different internally and externally and designed from the ground up as a new device, while sonoff basic and basic R2 was just a board revision maintaining the outside appearance.

and no, it is not "the" sonoff basic to users since everything in the past few years referred to the old design, this is the Sonoff Basic R3 and nothing else

PR #66, opened 2019-07-22
Add Sonoff DIY info

arendst:


blakadder:
I think Sonoff Basic R3 should be listed as a separate device since there's a big change in design and flashing process

arendst:
Agree.

I also just added the Sonoff DIY info to the wiki at https://github.com/arendst/Sonoff-Tasmota/wiki/Sonoff-DIY

PR #76, opened 2019-08-01
Formatting, Language, Article progression

meingraham:


blakadder:
why is this device entry being changed so often?

meingraham:
It's a complex device. I think the 1st pass was the template. Then two passes to start the guide and complete the guide. I made some changes because it lacked a bit of clarity. Also some links and references where missing or not the proper link. I think this will stabilize the entry now.

PR #77, opened 2019-08-01
GPIO/Component Table Render Fix

meingraham:
Apologies! For some reason I can not preview how the edits will render. When I click preview, it just shows me the edit differences and not the rendered view. I missed and extra line break before the table so the table did not render.


blakadder:
no problem

PR #85, opened 2019-08-15
Update Shelly Plug S - add temperature sensor

tobby88:
This adds the temperature sensor of the Shelly Plug S


blakadder:
Thank you

PR #90, opened 2019-08-21
Fixed template after 6.6.0.6 changes

DavidFW1960:


blakadder:
What's the point in deleting available GPIO's on this device?

DavidFW1960:
They aren't used or necessary in a default configuration mimicking the itead firmware. @meingraham when we were working through the 6.6.0.5 bug also suggested setting those to 0.

blakadder:
So you're removing user GPIO's on a device that is specifically made to use them via a jack plug??? What does the itead firmware have to do with anything?

DavidFW1960:
If I HAD a jack plug (or someone else did) they could just re-enable them. The config I propose just makes the plug work the same as the original firmware.

PR #98, opened 2019-09-01
Create kogeek_KLUP1_variant

eastabrooka:
Hi @blakadder, 
Thanks for running the Templates page. Your a lifesaver. 

Adding this pull request. 

After recently ordering 2x KLUP1 Plugs, and having them work fine with [This Profile](https://blakadder.github.io/templates/kogeek_KLUP1.html) 
I was surprised when I ordered a third, and it did not work out of the OTA Update process. 
I am offering this pull request to help someone else, if they find themselves in the same spot that I was.


blakadder:
Thanks for your contribution

PR #113, opened 2019-09-22
Update moeshouse_MS-104B

cHunter789:
added links to scripts in files because problem with display the code on site. Added info about original fw and version.


blakadder:
No need for this, it has been solved with a simpler solution

PR #127, opened 2019-10-06
Update template for SWB1

mikeg1130:
The template for SWB1 does not match what is on the Tasmota GitHub. Updated to match, and can confirm the new template works on my SWB1s flashed with tuya-convert.

https://github.com/arendst/Sonoff-Tasmota/wiki/SWB1-Smart-Power-Strip


blakadder:
Thanks

PR #139, opened 2019-10-18
122 is not a valid component number

mjoshd:
Valid component numbers are listed at:
https://github.com/arendst/Sonoff-Tasmota/wiki/Templates#components


blakadder:
There's been too many edits on this template creating problems for other users so I've merged the latest pull request for this

PR #142, opened 2019-10-19
Update AWP04L Template

CPad68:
It looks like the template was changed to {"NAME":"AWP04L","GPIO":[158,255,255,131,255,134,0,0,21,122,132,56,0],"FLAG":0,"BASE":18} which causes my unit to lock up and clear it's settings. I pulled this template from one of my working units (installed last weekend) and it is working now.

I've had several others run into the same issue and this change has worked for them. We are chatting on the DrZzzzs page on Facebook.


blakadder:
Seems there was a commit a few days ago that changed the template... Will merge this

PR #151, opened 2019-10-23
Update teckin_ss30.markdown

jrprinty:
Changed GPIO16 from Led3i (58) to Relay5 (25). This allows the USB charging ports to be turned on and off. There is not a 3rd LED.


blakadder:
Thanks

PR #161, opened 2019-11-02
Update arlec_PC189HA

Kanga-Who:
Update to the template as many users who have tried the previous configuration on the Home Assistant forums had issues with it working correctly, button not switching on/off etc.

This amended configuration is known to be working.


lukeoly:
I can confirm @Kanga-Who's changes work on the PC189HA

PR #176, opened 2019-11-11
Better configuration

Baldhor:
Top button blink blue when recevied an order
Bottom or Top button light red during movement, and stop immediatly when the position is reached.
Stop button is working perfectly (blink red when pressed).


Baldhor:
Sry I forgot to update the template itself !!! I will make a second pull request :(

PR #180, opened 2019-11-15
Update medion_life_S85225

RichardPillay:
The power button LED on GPIO13 was listed as 158 (LEDLinki), but I got better results after changing it to 56 (LED1i). Before the change, the LED lights up when the button is pressed physically, but of course you don't see it as your finger is on the button. After the change, the LED toggles to match the state of the relay.


RichardPillay:
Thanks a lot. I'll delete them now

PR #189, opened 2019-11-20
Update oil_diffuser_550ml

daaf84:
The light was going crazy, so changed this line.


blakadder:
what made it go crazy?

daaf84:
It did not respond in HA (switch went back off and diffuser was not responding). When I switched it fast on/off in HA, the diffuser began beeping like crazy. I had to pull the plug to make it stop.
Changed this line, now it works

daaf84:
I know now what went wrong; I changed diffuser to diffuser1 in HA, but not in the rules. This can be closed.

PR #193, opened 2019-11-21
Update tantan_WP3.markdown

dave-parsons:
This template was needed for mine to work - purchased 2019.


blakadder:
alt templates should be placed below the second `---` in the description filed or better yet, submitted as template v2 of the device with a new purchase link

PR #200, opened 2019-11-23
Update lohas_9W_E26_RGBW

ofields:
Fixed GPIO template


blakadder:
Is your bulb ordered from the same link? Seems to be bulb with completely different hardware inside

thehookup:
Just got a set of lohas bulbs from Amazon that need this updated template to work.

ofields:
Yes, confirmed my bulbs are from that exact link (Amazon has a banner on that page indicating I already ordered them once)

PR #202, opened 2019-11-26
Added alternate template for Slitinto NX-SM112 16A UK power monitoring sockets.

aso595:
Added third template I used to get my Slitinto sockets working. Also removed double quotes on the 2nd template as they don't appear to be required. Also added a couple of alternate product links.


aso595:
These approved changes appear to have been overwritten by the commits done today on Nov 27, 2019. The 3rd template I added was different to the other two already on the page.

PR #211, opened 2019-11-30
fix invalid template

thehaxxa:
Use Tuya TX instead of Serial TX, (makes the template work, currently it doesn't).


blakadder:
good job! thanks!

PR #218, opened 2019-12-02
Update wiki links; dpID table; Rules "line breaks"

meingraham:


blakadder:
languages syntax marking now supports ` ```console` like the docs

i've changed the template myself with the proper labels. tables don't look too hot in comments so I didn't include those

PR #221, opened 2019-12-03
Adds Button1

HenrikTo:
Adds possibility to toggle relay with Button1 (on GPIO 13)


blakadder:
you have two gpio's mapped to the same component, that is unacceptable

PR #224, opened 2019-12-04
Create Kogan_Smarthome_10W _Ambient_RGBW

knottyau75:
COnfig for yet another version of the Kogan Smart Home Ambient Bulb


blakadder:
that is not a properly filled out template. better use the Add Template link on the site

PR #239, opened 2019-12-10
Update template

mbelangergit:
From newly received units ordered on amazon.ca.
- Button was not on the same GPIO as in the database.
- GPIO13 had not effect on led as in the database. The led seems to toggle with relay whatever the config.
- Changed GPIO4 to LedLinki (personal preference)


blakadder:
please submit a new v2 template since the old template might still be used by old stock

PR #243, opened 2019-12-13
Update oil_diffuser_550ml

so3n:
fix formatting issue


blakadder:
thanks but there's some unnecessary edits, I've fixed the initial problem

PR #245, opened 2019-12-16
Update aisirer-JH-G01B

HemiBob:
In the original version, both the LED and relay were ON when the output (and HA) said it was OFF, and vice versa
Swapping the positions of 56 and 21 around fixes this.
However, the button on the device is not working so I suspect the "17" is in the wrong place - need further investigation.


HemiBob:
Now got the button to work correctly by moving "17" and changing FLAG to zero
Template should now be
{"NAME":"AISIRER","GPIO":[0,0,0,0,21,17,0,0,56,0,0,0,0],"FLAG":0,"BASE":18}

PR #251, opened 2019-12-21
Update aoycocr_U3S

mrvladis:
Update on ability to flash via OTA


blakadder:
Thank you for the contribution and happy holidays

PR #252, opened 2019-12-23
Update teckin_SP23

RoulstoneR:
It no longer uses ESP8266 Module and cannot be flashed.

Add - 23/12/2019 *** WARNING *** This model can now be supplied without power monitoring and it does not use the ESP8266 chip. This means it CANNOT be flashed with Tasmota by any method. See this blog post over at Scargill's Tech Blog https://tech.scargill.net/teckin-sp23-smart-socket/


blakadder:
Please add it to unsupported list instead https://templates.blakadder.com/new-unsupported.html

PR #253, opened 2019-12-23
Update moes-MS-105-1

apc300:


blakadder:
You're changing the entire template which means its a new device

apc300:
hi blakadder
would you like me to provide you with photos of the packaging and instructions for MoesHouse MS-105 Wifi Dimmer, bought from MoesHouse aliexpress?
i believe this is the listing for the correct device
thanks

blakadder:
yes but there's also the previous device which you're just deleting

apc300:
There is already a page for the previous device version https://github.com/blakadder/templates/blob/master/_templates/qs-wifi_D01_dimmer

blakadder:
and your point is? just submit a new template or don't

apc300:
could you elaborate please, i'm unsure why the device is not the same as the current template

PR #258, opened 2019-12-25
Update DS-102

ThomDietrich:
Simple additions based on own observations.


ThomDietrich:
@blakadder it's often preferred to avoid something like this to keep the holy git history clean:

grafik

That's why most repository maintainers would select the option "Squash and Merge" instead of "Create a merge commit", which would end in just two clean commits showing the PR title (you can even change the commit message). All the best! ;)

blakadder:
I prefer it this way

ThomDietrich:
Totally up to you of course. While you are the only maintainer of the project this method won't hurt much.

PR #260, opened 2019-12-27
Do not wrap gpio table content

ThomDietrich:
Avoid wrap on https://templates.blakadder.com/blitzwolf_SHP5.html

### DIFF
--- a/_sass/milligram.scss
+++ b/_sass/milligram.scss
@@ -552,6 +552,7 @@ ul {
 table {
   border-spacing: 0;
   width: 85%;
+  white-space: nowrap;
 }
 
 td,


blakadder:
that's a very blunt "solution"

ThomDietrich:
what's blunt about it? Neither the GPIO names nor the component descriptions will exceed certain length, hence this is totally fine.

What's not so cool about the PR is that it would affect all tables, which might not be desired. You need to add an `id` to the table which is currently missing. If you were not so quick to close the pull request I could have added that change 🤷‍♂

blakadder:
affecting all tables is the blunt part of it. 

what would a table id accomplish there?

ThomDietrich:
https://www.w3schools.com/html/html_id.asp

blakadder:
wow! didn't know that... thanks

PR #263, opened 2019-12-28
Update zemismart_KS-811_3gang

simbesh:
Fix broken image URL


blakadder:
link wasn't broken in the first place and your link is broken

PR #264, opened 2019-12-28
Update zemismart_KS-811_1gang

simbesh:
Fix broken image URL


blakadder:
not broken while yours is

PR #266, opened 2019-12-28
Update fcmila_7w

majorsl:
Change title to reflect that 9W version also works with this template.


blakadder:
Create a new template with links and images to that bulb

PR #276, opened 2019-12-31
Update WF-CS01_EU

akram:


blakadder:
yeah... that's not how that works

PR #278, opened 2019-12-31
Update naamasmart_KS602

majorsl:
Template didn't make use if the WiFi indicator LED at all. My change makes it blink when no connectivity and off when there is.  I updated the description to note that reversing the GPIO13 from LedLinki to LedLink will turn on the indicator always if wanted.  This depends if the user wants their switches glowing wifi indicators all over their home (I do not). Feel free to reverse that and the description if you feel it better.


blakadder:
all good, thanks!

PR #281, opened 2020-01-01
Added Italian seller and appropriate template

Magiweb:
Found to be very similar to .de configuration, but current measurement was incorrect. Adjusted and tested against known load.


blakadder:
Can you submit it as a new template instead?

PR #282, opened 2020-01-02
New file as requested

Magiweb:
Effectively the same plug as slitinto_NX-SM200 - just another derivative. 

Not sure if NX-SM2001 is inaccurate (extra "1" on the end) but that's the model number indicated on Amazon.it


blakadder:
all good now!

PR #283, opened 2020-01-02
updated to add template that is working for costco us

CanDoAnything:
I bought a bunch of these from Costco last year and all of them work w/ the template (which I found on https://github.com/ct-Open-Source/tuya-convert/issues/99 see comment at the end) 

The current template marked on this page only worked for the new bulbs I picked up from Costco yesterday.


CanDoAnything:
As i dig a little deeper, I have two variants of this bulb/model that were sold in the same packaging at different times:

**BPA800/RGBW/AG/2**
- Bought in 2018
- Works with:
  - 

**BPA800/RGBW/AG/2(P)**

PR #295, opened 2020-01-06
Add PWM5 to GPIO5

digitalec:
Light is not tunable white but has separate white control. By setting GPIO5 to PWM5, this enables the temperature slider which any adjustment switches from RGB to W.

### DIFF
--- a/_templates/nitebird_TT-WB4
+++ b/_templates/nitebird_TT-WB4
@@ -6,7 +6,7 @@ category: bulb
 standard: e26
 link: https://www.amazon.com/gp/product/B07MTYVSSV
 image: https://m.media-amazon.com/images/S/aplus-media/sc/b70090ee-c12e-4454-9a59-02721cb27415.__CR0,0,600,600_PT0_SX300_V1___.jpg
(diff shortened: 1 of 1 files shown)


blakadder:
If its not tunable white it can't have 5 pwm channels

digitalec:
Ah, I see. How else can you toggle from RGB back to white then? The Tuya app had the normal color wheel and then a separate “white” option to jump back.
    	
    	

    	
    
  




On Mon, Jan 6, 2020 at 7:20 AM -0500, "blakadder" <notifications@github.com> wrote:












If its not tunable white it can't have 5 pwm channels

blakadder:
https://tasmota.github.io/docs/#/Lights

digitalec:
Ah, thanks! Lots of good information there. I see 4Ch can have two toggle switches (I’m guessing Power + White). I’ll play around with this.
    	
    	

    	
    
  




On Mon, Jan 6, 2020 at 11:25 AM -0500, "blakadder" <notifications@github.com> wrote:












https://tasmota.github.io/docs/#/Lights

PR #302, opened 2020-01-10
Update and rename lsc_smart_connect_MR16 to lsc_smart_connect_GU10

pietpuk978:
The old version suggested its an MR16 bulb (12V with thin pins just like the G4 halogen)
But it is a GU10 bulb (230V thick pins with an extra locking thicker part at the bottom)
Also added the power (I have this bulb and it says 5 watt on the packaging)


blakadder:
MR16 is the shape of the bulb but is also used interchangeably for GU5.3 bipin. https://en.wikipedia.org/wiki/Multifaceted_reflector

PR #304, opened 2020-01-11
Correct LEDs in Torchstar LITEdge 2-in-1

gigem:
The LEDs should be the inverted variants.  I messed this up in my original submission.


blakadder:
thanks

PR #307, opened 2020-01-13
Updated Avatar AWP07L template

muhkuhgh:
Recently purchased them 4 through the provided link and tried the old template.. didn't work at all. Everything was mixed up, current and voltage after calibration voltage varied with load and current stayed constant etc.. it really took me some time but this template works perfectly fine for me. Allthough in comparion to my Gosund SP112 Plugs these one are veeeeeery slowly measuring..


blakadder:
moved to a v2 template

PR #310, opened 2020-01-15
Add HLW8012 based power monitoring to 35511

DelusionalAI:
Page also needs the "For more accurate energy consumption measurements this device requires power monitoring calibration." line but not sure if the site auto adds that.


blakadder:
it auto adds it if the gpio's are defined as one of the power monitoring components

PR #315, opened 2020-01-19
MS-105 Dimmer Unsupported

Xeyame:
Looks very similar, purchased from the link but is a fake clone.

### DIFF
--- a/_templates/qs-wifi_D01_dimmer
+++ b/_templates/qs-wifi_D01_dimmer
@@ -4,11 +4,12 @@ title: QS-WiFi-D01 150W
 category: switch
 type: Dimmer
 standard: global
-link: https://www.aliexpress.com/item/33010332202.html
+link: https://www.banggood.com/MoesHouse-DIY-Smart-WiFi-Light-LED-Dimmer-Switch-Smart-LifeTuya-APP-Remote-Control-12-Way-Switch-Works-With-Alexa-Echo-Google-Home-p-1509161.html
 image: https://user-images.githubusercontent.com/36734573/67634487-efaad080-f8bc-11e9-923f-4aa8babfd609.png
 template: '{"NAME":"WiFi-Dimmer","GPIO":[0,148,0,149,0,0,0,0,0,42,37,0,0],"FLAG":0,"BASE":18}' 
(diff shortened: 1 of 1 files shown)


blakadder:
The images you linked to clearly show an ESP module TYWE2S which is supported

Xeyame:
Well i feel very dumb now lol, i didn't notice the marking on this one being different from a relay board that i got from the same seller, with a module named "WA2", (but that relay was a way different one than the one i acutally bought, and i think is not esp.)
afbeelding

blakadder:
WA2 is indeed unsupported

blakadder:
WA2 is indeed a module with a nonESP chip

Xeyame:
Yea. Waiting for the dispute from aliexpress, as i bought a way different model. I'm curious what's under the tin so ill look into that soon, and i might be able to fit an esp-01 into the enclosure instead :)

corbolais:
> WA2 is indeed a module with a nonESP chip. by the looks of it this might be the one in the unsupported list

[This site](https://smartpets.en.made-in-china.com/product/ESzJbYnKVcRg/China-Esp-8285-Tuya-Tywe2s-Smart-WiFi-Module-Develop-Wireless-Networking-Intelligent-Module.html) states, the `TYWE2S` module is a "ESP 8285 TUYA TYWE2S" module. In the description, they then call it a "ESP8266EX". Well, pick your favourite one. Least common denominator would be, it's an ESP[82XX]. So, it *is not* a "nonESP". But the module looks somewhat like the `WA2` and the page popped up searching for `WA2`, so why should there be a nonESP under the hood (granted, the looks may very well decieving)? Any sources confirming this statement?

This `WA2` module is sitting inside those [smart switches](https://www.ebay.de/itm/2-4G-WiFi-Smart-Light-Switch-DIY-Circuit-Breaker-Module-Supports-for-Voice-Box/254456260792).

blakadder:
was there a point to be made there?

corbolais:
Well not any more. Technically, it will help users grasp the situation as it provides pointers and clarification.

Xeyame:
Might help random people lost on google yea

blakadder:
by stating WA2 is an esp based module when its clearly not?

Xeyame:
Oh yea sorry i was kind of lost in the wording of that comment, Yea WA2 is BK7231U and not anything ESP. However the link might help.

PR #317, opened 2020-01-19
Update after the 6.5.x release for SetOption14

MarcoVit91:
After the release of the update 6.5.x the rule SetOption14 has been dismissed and replaced with the "Interlock" rule.
You can check it here: https://indomus.it/news/hot-aggiornamento-6-5-per-sonoff-tasmota/


blakadder:
Thanks

PR #319, opened 2020-01-22
Update gosund-SW2

jasonehines:
I had issues with blinking when going over 85%. I adjusted the range and then ran into an issue where if you set the dimmer to 0 it would come back on; moving line 52 up into the previous statement fixed it. I've tested in the web interface, web console and Home Assistant. I also tested with an incandescent and a dimable led bulb.

### DIFF
--- a/_templates/gosund-SW2
+++ b/_templates/gosund-SW2
@@ -21,7 +21,7 @@ Minimum dimming level may vary by light fixture - adjust the ‘off’ value to 
 
 ```lua
 >D 100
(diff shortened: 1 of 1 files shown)


JCoder01:
@jasonehines I'm having an issue with the script where the `dimmer` command is sent with a parameter, it seems like none of the numeric variable are "working". Have you seen anything like this?

```14:44:36 f is ...d is ....
14:44:36 Script: performs "dimmer "
```

PR #332, opened 2020-01-30
Sonoff d1

kumy:
Please find some new devices and enhanced pictures for other models

### DIFF
--- a/_templates/emylo_XL9251WI
+++ b/_templates/emylo_XL9251WI
@@ -1,0 +1,15 @@
+---
+date_added: 2020-01-30
+title: eMylo XL9251WI
+category: switch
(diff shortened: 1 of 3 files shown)


kumy:
There is only one load output available to a standard user. do you mean we should also map Tx Rx as `user`? I ve attached pcb pictures may you cross check available feature please.

kumy:
+buzzer +led?
How should the rf receiver be mapped?

blakadder:
I don't know, I don't own the device. If I did I'd already post a template

kumy:
Didn't you closed it by mistake? ;)

blakadder:
Everything is added directly

PR #333, opened 2020-01-31
Add info about tasmota-lite.bin recommendation.

nagyrobi:


blakadder:
that's not an official recommendation

nagyrobi:
Yes it is. Check this out: https://github.com/arendst/Tasmota/wiki/Builds

> tasmota-basic.bin (bas) addresses a problem with some Sonoff Basics that lack sufficient electrical power (due to its poor design) to properly run the electronics. In this firmware variant, most sensors are disabled to minimize background tasks. This reduces the required CPU cycles in order to reduce power draw to allow the device to run reliably. 

And  tasmota-basic.bin was renamed later to tasmota-lite.bin, see changelog: https://github.com/arendst/Tasmota/releases/tag/v7.2.0

> Change tasmota-basic.bin and FIRMWARE_BASIC to tasmota-lite.bin and FIRMWARE_LITE

This is a welcome feature, I have 4 Sonoff Basics which are running the LITE build and run perfectly since. I'd recommend to add this note so that this information could propagate to masses.

blakadder:
That was the recommendation for the old builds running arduino core 2.4.2 and 2.3, we do not recommend installing -lite as the initial firmware, it is to be used in case you have issues

PR #335, opened 2020-01-31
flash: [tuya-convert, serial] - doesn't show anyithing

nagyrobi:
And also made the text more exact that people know that both tuya-convert and serial method are viable.


blakadder:
it doesn't show since its not yet implemented if both options are possible

PR #339, opened 2020-02-01
Different template for a new v2 model

tiagofreire-pt:
Different template for a new model mounted upside down, as the factory firmware works.


blakadder:
template2 is not a recognized field... will add myself

PR #351, opened 2020-02-08
Shelly RGBW2: Update template to dimmable RGB

caspermeijn:
I had some problems with the current template of Shelly RGBW2. The leds are not dimmable (they are with the original firmware) and the device would reboot a few minutes after loading the tempalate. These changed fixed my problems.

This includes the following changes:
- Added "Shelly" to the template name
- Changed not connected to None
- Change status led from relay to led
- Changed button2 to user, as this would reboot my device
- Use PWM instead of Relay to make the led strip dimmable


blakadder:
That old template was not the best solution indeed... Thanks

PR #358, opened 2020-02-16
Removing amazon link

declanshanaghy:
This bulb is not available on amazon, the on that is on amazon is https://templates.blakadder.com/lohas-ZN033.html

Note product # difference: LH-APP-W9-E26-4 vs LH-APPW9-E26-4


blakadder:
The manufacturer's link leads to the amazon link in the template so i'll stick with the manufacturer link for now altho I believe ZN001 is discontinued and replace with a newer version

PR #362, opened 2020-02-21
Update blitzwolf_SHP8

oxigenxservices:
Tamplate not function button.


blakadder:
what is this PR trying to accomplish?

blakadder:
Both templates are identical

PR #364, opened 2020-02-22
Add SwitchMode0-like script

antonio-fiol:
This device is intended to work like Switch Mode 0 (working like a 3-way switch, toggling the output when the input is toggled, or when toggled via wifi). Neither of the provided scripts did that, so I tried to adapt them. After some testing and removing glitches, I am fairly confident with the provided script, but feedback is more than welcome.


antonio-fiol:
By the way, I am sure that the script can be written in a better way... this is my first script.

PR #368, opened 2020-02-23
Update moes-MS-104B

cHunter789:
Sleep 50 is somehow not working good for timers


ascillato:
Hi,

> Attention: I have found that the script (at least for switchmode 1) is somehow depended from sleep setting. Swichmode 1 script is tested and should be used with sleep 200 or 250. 

Are you sure? have you tested with default `sleep 50` ?

Checking Tasmota code, `switchmode` is not dependant on `sleep`. switchmode is just what will be the behaviour of that particular switch.

cHunter789:
Ok hold it. I have found something else. Just cancel this pull.

PR #372, opened 2020-02-26
Update treatlife_SS02

majorsl:
URL points to a newer version of this switch which cannot be flash with tuya-convert (not the ESP chip) which I found out the hard way.  The link I put in goes to the older version of the switch.  The replaced link should probably go with a warning in the incompatible devices section.


blakadder:
The link you've changed is the manufacturer link not the purchase link

majorsl:
Apologies for not getting back sooner.  Proper URL is:

https://www.amazon.com/gp/product/B07ZSQ8R85/ref=ppx_yo_dt_b_asin_title_o06_s00?ie=UTF8&psc=1

PR #391, opened 2020-03-10
Update 2nice-UP111

diegoherranz:
- LedLink and Led1 swapped. The blue LED is on GPIO1 which works better as LedLink and the red LED is on GPIO13 which works better as Led1.
- Both LEDs set as active low (LedLinki and Led1i) so that the red LED is on when the plug is on, and the blue LED is on when indicating wireless activity, issues, etc.

Thanks!


diegoherranz:
Thanks!

PR #396, opened 2020-03-14
Wrong info.

pesor:
Sorry, after a couple of more tries, it actually flashed. 

So it is Tasmota compatible, and should in stead be added to that list.


blakadder:
there's 0 context here

PR #397, opened 2020-03-15
Updating TanTan WP3

CraWleRLT:
Tried using this, however nothing worked. Tried [Gosund WP3 Plug template](https://github.com/blakadder/templates/blob/master/_templates/gosund_wp3) afterwards and that worked like a charm, including:
* Physical button clicks toggling the power on/off
* Tasmona toggle turning the power on/off
* Light indicating whether the power is on/off
* Blinking red light indicating MQTT did not connect

### DIFF
--- a/_templates/tantan_WP3
+++ b/_templates/tantan_WP3
@@ -4,9 +4,10 @@ title: TanTan WP3
 category: plug
 type: Plug
 standard: us
-link: https://www.amazon.com/TanTan-Wireless-Required-Anywhere-Upgraded/dp/B071VYFJRL
+flash: tuya-convert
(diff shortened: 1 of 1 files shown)


blakadder:
I notice you bought the device from a different link so its understandable the template might not work. you have a different revision which works with another template with the same model name

CraWleRLT:
@blakadder, I assume you suggest adding a new template instead then, right? If yes, should I add it with `Reference Number/FCCID` as `T10AL120V` or `B12` or something other? See the bakc of the device:
tantan_wp3

blakadder:
yes, create a new TanTan WP3 v2 template

PR #418, opened 2020-04-05
Added Arilux LC06 template for CW/WW mode

SilverFire:
Related to https://github.com/arendst/Tasmota/issues/2117#issuecomment-609477652


blakadder:
That doesn't need a template since you can change gpio's to your liking

SilverFire:
The default template of the module does not work with CW/WW at all: only WW is dimming and nothing else. There is an issue https://github.com/arendst/Tasmota/issues/2117#issuecomment-609477652 where people are monkeying around for 2 years trying to make this module work with CW/WW.

Lucky me adjusted GPIOs in a way they work and would like to share knowledge with new Tasmota users. I do not introduce a new template since the device is the same, instead, I show how to use the device with CW/WW strip.

@blakadder Could you re-evaluate the PR, please? Maybe it makes sense to you with this description?

blakadder:
monkeying for 2 years?
if you need only two channels you enable only two PWMs, if you need RGB you enable 3 PWMs etc... sure, that was not easily done before templates but with templates its a simple and easy process 
i'll include the PR since its apparently not as straightforward for some

PR #439, opened 2020-04-18
Update hugoai_awp02l-n

finnzz:
Assuming that this is the same plug as the HugoAI version, I would recommend a slight change to the template.

At least on the HugoAi version, setting GPIO00 to LED2i (57) prevents the power LED from turning on when the plug is toggled on.  Setting GPIO00 to LED3i (58) instead, gives you both a power LED when toggled on and the MQTT heartbeat blink when there is no connection to the MQTT server (but only when the plug is in the off state).

### DIFF
--- a/_templates/hugoai_awp02l-n
+++ b/_templates/hugoai_awp02l-n
@@ -11,3 +11,13 @@ link2:
 ---
 
 
+
+Title: AWP02L-N (FCCID:2ANJP-AWP02LN)
+Category: plug
+Standard: US
+Template: {"NAME":"AWP02L-N","GPIO":[58,0,56,0,0,0,0,0,0,17,0,21,0],"FLAG":1,"BASE":18}
+Link: https://www.amazon.com/HUGOAI-Outlet-Required-Control-Devices/dp/B07D9G92BF
+Image: https://images-na.ssl-images-amazon.com/images/I/51Rj-KwLo1L._AC_SL1500_.jpg
+
+
+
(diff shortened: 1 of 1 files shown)


blakadder:
And what happened to Led1 ?

finnzz:
Ah good call setting GPIO00 to LED1 gives the same exact behavior as setting it to LED3/LED3i.  I get a power light when toggled on, and when toggled off and no MTQQ server, I get the slow heartbeat blink.

So for simplicity, changing GPIO00 to LED1 (52).

{"NAME":"AWP02L-N","GPIO":[52,0,56,0,0,0,0,0,0,17,0,21,0],"FLAG":1,"BASE":18}

I'm not sure if there is a configuration that gives the MQTT heartbeat while the power is toggled on as well, but this template is still better for the HUGOAI varient of AWP02L-N I have.

blakadder:
this template isn't very different from the original one and neither is using LedLink to properly display network status

finnzz:
Yeah it's pretty much identical, except the Power on LED doesnt work with the original template.

Now after playing with the settings again, I noticed that GPIO00 is not even required.  Setting it to none gives me a functional power LED On light and the MQTT heartbeat (assuming this is what you mean when you say network status). I'm not sure what the purpose of GPIO00 is in this case.

This is the first template I had to modify.  If you'd like me to try anything else let me know.  

{"NAME":"AWP02L-N","GPIO":[0,0,56,0,0,0,0,0,0,17,0,21,0],"FLAG":1,"BASE":18}

PR #440, opened 2020-04-18
Updated Arlec GLD060HA Template

jrangama:
This light has only warm white colour. Removed the colour temperature GPIO from the template.


blakadder:
thanks

PR #443, opened 2020-04-19
added power monitoring pins.

Sthopeless:


Sthopeless:
@guardiande it was the only power metering config that give me any results on the webpage but they seemed not calibrated. 
There's a change is not the correct sensor/pinout.

I haven't opened the device yet but the screws are triangle shape and are under the silicone pads/feets. If I ever get to open it I will post some pics here

PR #448, opened 2020-04-21
Update WF-CS02_EU

Krawollo:
Control background light of all three buttons with Relay4 on GPIO1.
Used with WF-CS02 from Maxico


blakadder:
Why not use Led for that?

Krawollo:
With Relay4 it is possible to switch background light of all 3 buttons it on and off. Even with voice commands from Alexa.

blakadder:
Sure but that's not the correct way to configure

PR #453, opened 2020-04-25
add Script

Bettman66:
To be controlled completely with script language


Bettman66:
Hello Blakadder,
the formatting of the script is not nice. Did I do something wrong ?

blakadder:
probably

PR #456, opened 2020-04-28
Update lohas-ZN022

gdreelin:
Button was in the wrong position. Template should look like this {"NAME":"LOHAS M5-022","GPIO":[0,0,0,0,38,37,0,0,17,39,0,0,0],"FLAG":0,"BASE":18}

### DIFF
--- a/_templates/lohas-ZN022
+++ b/_templates/lohas-ZN022
@@ -3,7 +3,7 @@ date_added: 2020-03-26
 title: Lohas ZN022 5m RGBW
 model: LH-Light Strip-RGBW
 image: /assets/images/lohas_M5-022.jpg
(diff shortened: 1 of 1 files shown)


gdreelin:
I need to make this change to this one.

{"NAME":"LOHAS M5-022","GPIO":[0,0,0,0,38,37,0,0,17,39,0,0,0],"FLAG":0,"BASE":18}

blakadder:
maybe i'm blind but i don't see the difference between your two comments, please edit the PR to the correct template instead of posting it in the comments

gdreelin:
> 
> 
> maybe i'm blind but i don't see the difference between your two comments, please edit the PR to the correct template instead of posting it in the comments

Sorry new to this stuff. I did the change but it said I did not have access to do so. Both comments are the right one now. It said I have to do a pull request but I am new to Github so was not sure if I was doing it right.

PR #466, opened 2020-05-07
Update etersky_WF-CS01

fennec622:


robitherobo:
hey @fennec622 @blakadder 
I'm new to all this and I'm seriously not in my element.. sorry if I'm asking something dumb af 
how do I add these switch to Homebridge? only way I managed to do it was basic on/off not  up/stop/down as normal behaviour. Can't find a plugin for it and it's not loading from either smart life or Tuya's apps 
thank you in advance

PR #472, opened 2020-05-12
Changed analog type

effelle:
The Witty cloud has an LDR, hence should use Analog Light (3).


effelle:
@blakadder Theo wants it as pure analog instead, perhaps is better just adding a second possible template instead of changing the default.

blakadder:
Sure, I can have 2 templates now

PR #473, opened 2020-05-13
mod: uses esp8266 and can be used w/o scripts

corbolais:


blakadder:
not the same device anymore

corbolais:
True.

In the light of future users looking for support may profit from this note. They might consider modifying it instead of discarding the device.

HTH

PR #475, opened 2020-05-13
Update bneta_GU10

NTRNSTR:
Was unable to get color working on this bulb with existing template, this new config seems to work with both white and color being controllable and functioning as expected. There is no slider for white but when set using the console it worksas it should.


blakadder:
thanks...

there is no white slider when only single channel white is configured

PR #477, opened 2020-05-16
Update arlec_PBCB3HA

insanemal:
Just updated with most recent experience based on not finding this template before just making my own based on other Arlec templates


blakadder:
yes, your configuration is the proper one. thanks!

PR #486, opened 2020-05-20
Update Deta 4 Gang

jimtng:
The button order is clock-wise. With the status light is on the top, it starts from (1) bottom left, (2) top left, (3) top right, (4) bottom right. This also corresponds to the terminal labels on the mounting block.


blakadder:
thank you

PR #487, opened 2020-05-21
[bauhn_ASPU-1019] Merge two relays into one

jimtng:
This device only has one outlet, and the previous settings created two relays, as a result, the MQTT RESULT contains POWER1 instead of POWER. There is no need to be able to control each relay independently. By setting both GPIO4 and GPIO5 to `relay1` it removes the need for the "power2 on boot hack" and makes the MQTT result cleaner. Therefore this will appear as a single relay device, without having to deal with POWER1 / POWER2.

Also fixed typo in the template name

### DIFF
--- a/_templates/bauhn_ASPU-1019
+++ b/_templates/bauhn_ASPU-1019
@@ -6,16 +6,10 @@ type: Plug
 standard: au
 mlink: https://bauhn.com.au/wp-content/uploads/2020/02/ASPU-1019-Smart-Wall-Plug-UM-v8.pdf
 image: /assets/images/bauhn_ASPU-1019.jpg
-template: '{"NAME":"Buahn Smart Pl","GPIO":[0,0,0,0,21,22,0,0,0,56,17,0,0],"FLAG":0,"BASE":18}' 
+template: '{"NAME":"Bauhn Smart Pl","GPIO":[0,0,0,0,21,21,0,0,0,56,17,0,0],"FLAG":0,"BASE":18}' 
 link2: 
 ---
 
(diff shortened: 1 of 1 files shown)


blakadder:
Two Relay1's makes even less sense and is definitely worse.

If the device has only one controllable power output then you only need one Relay1, don't know what the Relay2 was exactly. Define only one Relay1 on the pin that actually controls a relay and that should be it

jimtng:
For the output to be turned on, both gpio4 and gpio5 need to be set to relay1 as per this PR.
If I defined relay1 on only one gpio, i.e. either just on gpio4 or just on gpio5, it won't work.

blakadder:
Weird hardware design choice but ok ;)

jimtng:
Please revert this. I apologise, I was wrong. It appears that assigning relay1 to multiple gpio isn't supported by Tasmota.

deswong:
Assigning Relay1 to multiple GPIO's definitely is not supported by Tasmota.
I assigned Relay1 and Relay2 in my pull request, but there are no areas to add any weird configuration notes. 
What I did after assigning Relay1 and Relay2 was to have a rule - when Relay1 was powered on to copy to Relay2, and vice versa. 
It is a really weird design - I suspect they have done this in the newer models as a failsafe for hacking - that it would only allow either Neutral or Active, but not both, unless you knew that you had to run the two together.
I was terribly frustrated with this design - to the point that the next one I Tasmota I might just hard short out the second relay so it is less hassle to get it up and running!

jimtng:
@deswong You could add the note at the end of the page.

PR #497, opened 2020-05-30
Added another simple way to use this module.

rokrodic:
Another way to do the same thing with some changes in the source code of Tasmota. No need to use scripts. Works flowlessly. Clear instructions on my homepage and in the Youtube video.


blakadder:
thank you very much

PR #499, opened 2020-05-31
Update WF-DS01

esackbauer:
Missing ledtable command, mine was on by default causing dimming only on the last 30% and leaving 70% unused.


esackbauer:
See here for more info:
https://github.com/arendst/Tasmota/issues/7997

PR #500, opened 2020-05-31
Change picture URLs

roberttco:
Change the picture URLs to fully qualified URLs and properly terminate the markdown with a closing paren.  @blakadder  - obviously with this change I am assuming that the file parser understands the absolute URL and the image markdown.


blakadder:
You just didn't close brackets to fulfill markdown syntax

PR #503, opened 2020-06-02
Update Bauhn ASPU-1019

deswong:
There is a second relay on GPIO5 that also needs to be switched on/off to control output


blakadder:
can you please reconcile this with #487

jimtng:
It was my bad PR that broke this, I'm sorry. Here is the original template:

```
template: '{"NAME":"Buahn Smart Pl","GPIO":[0,0,0,0,21,22,0,0,0,56,17,0,0],"FLAG":0,"BASE":18}' 
```
There was also a note in the page that explains it:


> This device uses two relays, so use the following commands to setup a rule to trigger the second relay when the first is triggered
>
> ```console
> rule1 on power1#state do power2 %value% endon
> rule1 1
> ```

jimtng:
Alternatively the rule could be:

```
rule1 on system#boot do power2 on endon
rule1 1
```

This will turn relay2 always on and you just control power1. This will result in only one sound of relay clicking instead of two.

PR #528, opened 2020-06-17
updated LED assignments

annbob:


blakadder:
you've turned led's into Relayi ?!?

annbob:
Sorry for the misinformation. 
When I first created this template, I didn't know as much about the workings of Tasmota. Now I know a little more. When I first set this template up I got the LED's to function correctly by making them Relays. This method looks like it may have been a workaround for older firmware. The new GPIO assignment more accurately describes the unit.
The plug has 2 outlets and 2 LED's. No dedicated link LED. With GPIO3 set to 157 (LEDLink) both LED's are blue or blinking blue for status of WIFI/MQTT. When power is turned on to an outlet the corresponding LED turns orange.
GPIO12 is LED linked to Relay1.
GPIO14 is LED linked to Relay2.
Hope this clears some things up.
I'm also going to submit another PULL REQUEST that addresses the Rules. These changes to the GPIO and the newer firmware broke the Rules that were written.

PR #535, opened 2020-06-19
Update QS-WIFI-RGBCW

emilalbuaxia:
I've had time to play with the device, so i managed to get all the 5 channels and button working.
So, i've updated the page and added different kind of templates, depending on what type of led strip is used.


blakadder:
great job! thanks!

PR #544, opened 2020-06-26
Update winpro_next for latest Tasmota

raghavrithvik:


shresth45:
Hi. Wondering how did you flash Tasmota onto the tubelight. Bought one this month and tried the tuya convert method. It resulted in bricking the device

PR #549, opened 2020-06-28
Update bauhn_ASPU-1019

SuperEvenSteven:
The previous solution left the LED on regardless of whether the plug was in the on/off state. This fixes that plus corrects the previous boot workaround by syncing the second relay to the first one. This mimics the original behavior of the orginal Tuya firmware.


blakadder:
This plug is quite tricky :D

PR #550, opened 2020-06-29
Update LVWIT A70 12 correction

curlyel:
Warm white/cold white channel seem to be twisted in the current template. Proposing changing it.


curlyel:
Mmmh. The PR does not really solve the issue with the LVWIT. GPIO5 is the brightness for white LED and GPIO13 controls CT. So we have likely a PWM-CT here. Not sure how to solve it currently with just updating the template...

https://tasmota.github.io/docs/Lights/#pwm-ct

curlyel:
Replaced by: https://github.com/blakadder/templates/pull/559

PR #557, opened 2020-06-30
Prevent power trigger using Backlog

curzon01:
Backlog in Rule1 issues a possible short power switch on when using the stop button (see https://github.com/arendst/Tasmota/issues/8707)


curzon01:
PR already done by Zer0x00

PR #559, opened 2020-06-30
Update Template for LVWIT 12W RGBCCT Bulb

curlyel:
PWM5 (GPIO13) controls CT. Need to set to inverted mode (PWM5i) and additional setting of "setoption92 1" to enable PWM CT.
Should work since v8.2.0.5, tested with v8.3.1.6.

Don't know how to put the remark into the template file :-(
Should be shown in the template repository, otherwise the commands CT, dimmer and white will not work correctly.


blakadder:
done

PR #563, opened 2020-07-02
Update proficook_PC-WKS_1167

NovalisVonHardenberg:
I added all relays to WebUI.


blakadder:
That will create excessive flash writes and shorten the lifetime of the device for no real benefit

PR #574, opened 2020-07-08
Create wipro_garnet_NS9100_900lm

ChaitanyaPramod:
This is slightly different from NS9100 810lm, hence creating a new template. Alternatively could be merged into the same file as alternate template


blakadder:
will add as alternate since the buying link is also the same

PR #575, opened 2020-07-09
Update Moko Switch: remove power monitoring

TroyFernandes:
Bought this switch recently and it doesn't have any power monitoring. Also this template enables the LEDs


blakadder:
thank you

PR #577, opened 2020-07-11
Alternative template

aussiebyrd:
Added alternative template which defaults to LEDs off when switches are off. Consistent with template for 1-gang version.


blakadder:
If that switch has led's for each relay already I would recommend changing that led to ledlink instead

aussiebyrd:
Yep, one LED per "switch" (relay). I just tested ledlinki, and it seems to work fine. As such should both of the templates be edited to refelct the options? i.e. LedLinki and LedLink respectively? If so, I can jump in and update for the 1-gang and 3-gang versions of this as well.

aussiebyrd:
Updated both templates now.

PR #584, opened 2020-07-13
Update kogan-KASPEMHA

jamieparris89:
Updated GPIO0 from Button1 (17) to Switch1n (82) to make the button functional as it was not switching the device on/off when pressed.


jamieparris89:
Nevermind. I think I was having a different issue. Problem only affecting 2 of 4 sockets.

PR #587, opened 2020-07-18
Blitzwolf BW-SS4 one-gang / two-gang

Carko:
Blitzwolf BW-SS4 comes in "one gang" and "two gang" versions. 

Since everything else is the same, including the buy links, I have added the other one as an alternative template instead of a new device.


blakadder:
thanks

PR #589, opened 2020-07-21
Update martin_jerry_ST01-1P

qwertangel:
same as https://templates.blakadder.com/martin-jerry-3-way-ST01.html .


qwertangel:
same as https://templates.blakadder.com/martin-jerry-3-way-ST01.html. so either merge this commit or remove this switch completely from repo.

blakadder:
deleted the file

PR #593, opened 2020-07-24
Update avatar_ALS11L

letroll:


blakadder:
i'm sorry... what is this update about? what does "firmware before 1.0.2" mean"

letroll:
Tuya-Convert seems no longer able to flash Tasmota firmware on these device when their original tuya firmware is updated in version 1.0.2 with tuya mobile application.

PR #600, opened 2020-07-30
Clarify this is a Moes MS-104B clone

halfbakery:


blakadder:
Is the clone link just to link to relevant commands for setting switchdebounce properly?

halfbakery:
I mean the whole PCB layout of the BlitzWolf SS5 2 gang seems to be the same. The version markings are different (mine is QL-0362 REV:A), but all of the components are at the same position and same type/value. So the whole board is just a rebranded MOES MS-104B with a BlitzWolf logo. That's why I just wanted to put a reference to the moes page.

blakadder:
Take note there's a 1 gang and a 2 gang version too

blakadder:
I'm also a bit confused as how exactly the switchdebounce value of 69 relates to the number of pulses? 

Can you take a look at https://tasmota.github.io/docs/Buttons-and-Switches/#ac-frequency-detection-switch and help make it more understandable to a layperson

halfbakery:
Please change the counters to switches as you did in the moes entry, and put a link that it's a rebranded MOES MS-104B.

PR #601, opened 2020-07-30
Document the new AC detection mode

halfbakery:


blakadder:
Thanks. I took the liberty of making your way the default one since its vastly superior to anything else

halfbakery:
Thanks. I fully trust my solution, but I was not as brave as you :-)

PR #604, opened 2020-07-30
Update Connect SmartHome product range

mechanarchy:
* Add two new templates for 10W RGBW bulbs
* Rename one template file that was inconsistent - **this will break the old URL**
* Reorder YAML fields for consistency
* Update product names for consistency


blakadder:
while trying to introduce consistency you introduced doubling in product names, please remove all color references from `title:`

mechanarchy:
Ahh thanks for the pickup, I didn't realise the color gets included in the title as part of the template. I have rectified this now.

PR #607, opened 2020-08-01
Update gosund_KS-602S

idumzaes:
added corrected template as the first two did not work well.


idumzaes:
I suggest removing template-alt as it did not work at all.

PR #618, opened 2020-08-13
Update lumiman_LM530-E26

reven:
Added note that newer revisions no longer use ESP8266 chip and are therefore incompatible with tasmota.


blakadder:
thanks. i added the warning across the board for all lumiman products instead

PR #622, opened 2020-08-18
Fix Heyvalue HLT-331

declanshanaghy:
The given template did not work properly. 
Issues with old template
* Only 3 of the relays were controllable and they were not using inverted relays. 
* switch was not mapped to the correct GPIO and should have been mapped as a button. 
* One of the LEDs was not mapped at all


blakadder:
thanks

PR #653, opened 2020-09-05
amazon is delivering WB4 and is not compatible

antoniomolram:
Amazon links removed to avoid more people to buy WB4 thinking that are WB3...


blakadder:
thanks

davidebeatrici:
Is arendst/Tasmota#15879 the WB4 you were referring to?

PR #656, opened 2020-09-06
Update lenovo_SE-341A

wildflower:
added my experience of trying to open the case


blakadder:
Why is that relevant? SMD components can be soldered back on

PR #665, opened 2020-09-14
Update brilliantsmart_20967

nmbradford:
Add info about tuya-convert and physical flash not possible.


blakadder:
that's unfortunate, thanks for the info!

PR #666, opened 2020-09-14
Update merkury-MI-BW904-999W

tduffy83:
Updated to template that works, previous template gives no white output, only rgb.

### DIFF
--- a/_templates/merkury-MI-BW904-999W
+++ b/_templates/merkury-MI-BW904-999W
@@ -5,8 +5,8 @@ category: bulb
 standard: e26
 link: https://www.walmart.com/ip/Merkury-Innovations-A21-Smart-Light-Bulb-75W-Color-LED-1-Pack/254063201
 image: https://i5.walmartimages.com/asr/e15b4ae9-a1b0-4a6b-a93a-b818ee4858f2_1.1eeae0645b5aae960d7c57f7689c242a.jpeg
-template: '{"NAME":"MI-BW904-999W","GPIO":[0,0,0,0,140,37,0,0,0,142,141,0,0],"FLAG":0,"BASE":18}'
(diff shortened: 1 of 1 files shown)


blakadder:
Your template is for a 5 channel bulb and this is a 4 channel RGBW one

tduffy83:
The current template does not work with this bulb.  The white channel does not work, it only provides RGB.  The template I provide works fully, you can use the white channel.

tduffy83:
Did some tests and you need the two PWM channels for this model bulb to work with tasmota.  I agree it should only be 4 channels, yet the white channel simply will not work otherwise.  I am testing on the actual hardware, model number matches MI-BW904-999W.

blakadder:
What kind of tests exactly?

tduffy83:
Tests with only using one PWM in GPIO5 and GPIO12, respectively.  With one PWM in GPIO5 you get what I describe, no white channel.  With one in GPIO12 you get only white output, but color sliders that only make the bulb flicker slightly when using them.

tduffy83:
I will admit this causes issues in home assistant and I imagine other home automation software, as the bulb gets set as a RGBCCT light and a color temperature slider is presented.  It's easily fixed in home assistant with a supported_features over ride.

blakadder:
I hope you are aware 4 channel lights do not offer a white slider as documented in Tasmota docs

tduffy83:
Interesting, in that case there is no way to turn on the white channel via the tasmota web interface if this device is configured as a 4 channel device.  I did test the original template and I can confirm that the bulb is added correctly in  home assistant and home assistant works correctly (can turn the white channel on).  To me however, this is not ideal as I can forseee a future where I want to change the bulb via the tasmota web interface and will not be able to turn the white channel on.

tduffy83:
Ok so I found the documentation and it appears my ignorance of how the software is intentionally designed in a not so intuitive way led me to start this issue.  Why does tasmota not include a way to turn the white led on in the webui?  Doesn't need to be a slider, could just be a button.  Either way, I'm closing this as it appears the original template is the correct way to do things.

PR #671, opened 2020-09-21
Update feit_electric-OM60RGBWCAAG

chrisswanda:
I purchased these bulbs from Menards on Sep 21, 2020 and this config worked for me.


blakadder:
thanks

PR #682, opened 2020-09-30
Update treatlife_DS03

boybert:
Modified the (extremely helpful) example Home Assistant YAML to include the section of the configuration that the code applies to (i.e. "fan" and "light"), and included some additional example Tasmota Console readouts to help determine if you are successfully communicating with the MCU.  Added clarification that you may not receive MCU heartbeat unless the switch is assembled/installed/connected to mains.


boybert:
Thanks!

PR #685, opened 2020-10-01
Update Teckin SR40 image URL

JeffResc:
Updates the Teckin SR40 image from [this](https://templates.blakadder.com/assets/images/teckin_RF-SR40-US.jpg) to [this](https://templates.blakadder.com/assets/images/teckin_SR40.jpg) as the old one was a 404. Seems like the new image is the same product, as shown from [it's Amazon page image](https://www.amazon.com/dp/B07RFPSZLG?tag=blakaddertemp-20&pldnSite=1).


blakadder:
thanks

PR #688, opened 2020-10-03
Update Digoo DG-ZXD21 battery config

zegerk:
Added unit of measurement and unique id


blakadder:
why unique id?

if you have two of the digoo sensors the id will clash

zegerk:
Sometimes HA requires unique id - perhaps indicate to the user it is an example template and use a different id (and name) if you have multiple devices?

blakadder:
>  Sometimes HA requires unique id - perhaps indicate to the user it is an example template and use a different id (and name) if you have multiple devices?

no guessing please, if you're not sure what it does don't add it

PR #692, opened 2020-10-03
Update wyze_WLPA19

takelley1:
Add warning that Tuya-convert not longer works with device.


blakadder:
template already states serial flash not tuya-convert

PR #694, opened 2020-10-03
Update alfawise_PE1004

lordnaicon:
Strommessung


blakadder:
sorry, new template format is not yet supported until its officially released

PR #702, opened 2020-10-06
Add DP10 serial flashing note

cmccambridge:
I'm not sure whether these details are ordinarily included with a template entry, so no worries if you'd prefer not to add this type of data here and prefer to decline this PR!

I just thought I should pass them along, as the details would have saved me some research time this evening 😄


blakadder:
quite relevant, thanks!

cmccambridge:
Thanks @blakadder ! Very much appreciate your work on maintaining this database!

PR #704, opened 2020-10-06
Update lonsonho_X803A

ovidiu-vas:
Fix GPIO13; add relay for toggling background light.

### DIFF
--- a/_templates/lonsonho_X803A
+++ b/_templates/lonsonho_X803A
@@ -3,7 +3,7 @@ date_added: 2020-05-16
 title: Lonsonho  3 Gang
 model: X803A
 image: /assets/images/lonsonho_X803A.jpg
-template: '{"NAME":"Lonsonho X803A","GPIO":[0,0,0,18,17,19,0,0,21,29,22,23,0],"FLAG":0,"BASE":18}' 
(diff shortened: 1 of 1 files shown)


blakadder:
why do you have relay4i ???

ovidiu-vas:
That's a personal preference. I prefer to start with the background lights off but it can be set to Relay4 (24) to turn them on by default. I wrote this in the commit below the template code.

blakadder:
your personal preference is wrong because that device doesn't have a fourth relay

ovidiu-vas:
Please read the comment added in the commit:

**Relay 4 toggles the green LEDs on all buttons. To turn on the green LEDs from the start set Relay 4 to "Relay4 (24)".**

The 4th relay is not an actual physical relay. It's set as a relay so that you can see it in Tasmota as a toggle and switch on/off  the background LEDs.

blakadder:
Yes I understood that part. And it is still wrong. The fact it can be done and it works for you doesn't mean it should be done and it works for everyone else.

ovidiu-vas:
What do you mean? It works for everyone. The current template is bugged and forces the first relay to inverted because of how the second relay is done.

blakadder:
Then please submit a template without imaginary relays

PR #718, opened 2020-10-14
Update blitzwolf_BW-LT21

BugsonPL:
There is no PWM5 channel in this device.
If one wants to use PWM4 channel from WEB interface one should use:
-SetOption37 128 (split RGB and W channel)
or
-SetOption106 1 (enable virtual PWM5 channel from RGB)


blakadder:
thanks

PR #726, opened 2020-10-20
Update treatlife_DP10

mcdermittd:
Added guidance for adjusting the DimmerRange to accommodate common use case for this outdoor dimmer plug.


mcdermittd:
Added a short not about adjusting the dimmer range in order to accommodate common use case.

PR #735, opened 2020-10-27
updating globe_34207 template with costco.ca model change

cturra:
adding a note to share that the recent model number cannot be flashed with tasmota. also made minor updates to page formatting.


blakadder:
thanks

PR #736, opened 2020-10-28
Update mimoodz_A21

digitalec:
Previous settings were useful for only initial GPIO probe. GPIO5 & 12 are PWM1 (37) & PWM2 (38), respectively.


blakadder:
thanks for the fix!

PR #738, opened 2020-10-30
Update bn-link_BNC-60_U133TJ

akseidel:
Corrected some template parameters. Added construction and wiring access details.

### DIFF
--- a/_templates/bn-link_BNC-60_U133TJ
+++ b/_templates/bn-link_BNC-60_U133TJ
@@ -7,7 +7,28 @@ type: Plug
 standard: us
 link: https://www.amazon.com/dp/B07CVPKD8Z/
 image: /assets/images/bn-link_BNC-60_U133TJ.jpg
-template: '{"NAME":"BNC-60/U133TJ","GPIO":[0,56,0,17,134,132,0,0,131,57,21,0,0],"FLAG":0,"BASE":18}' 
-link2: 
-NOTE: As of October (prime day) 2020 the 4 packs of these have an updated firmware that can't be flashed with Tuya-Convert v2
+template: '{"NAME":"BNC-60/U133TJ","GPIO":[0,52,0,17,134,132,0,0,131,53,21,0,0],"FLAG":0,"BASE":18}' 
+link2:
 ---
+Uses a TYWE2S daughter board, which makes it a Tuya device but ...
+As of October 2020 the 4 packs of these have an updated firmware
+that can't be flashed with Tuya-Convert v2. The same applies to the 2 pack. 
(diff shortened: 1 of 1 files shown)


blakadder:
are those led colors same as default firmware behaviour?

akseidel:
I do not know for sure because I did not want to run the device with the vendor’s firmware, so I used the parameter template as it was originally published. I suspected from the start the red LED colors were wrong.

The vendor’s instructions supplied in the packaging did not describe at all what to expect for operation. I did look online for videos and pictures of the device in action. I found only one image that showed the led in blue. That color is what I would expect is a plausible “in operation” meaning.

This device’s TWE2S board has 11 connection points on its edge. I believe two are not connected to the main PCB board. These are RST (IC Pin 32 EXT_RSTB) and AD (IC Pin 6 TOUT) (parameter ADC0). All the template parameters made sense with the remaining available connections. Therefore there are no unused parameters to go with unused connection points that might be controlling a third led color some way. 

I did not try reversing the targets for the two led controlling parameters out of laziness because my initial thought to change inverse to non-inverse resulted in colors that made better sense to me. In hindsight perhaps I should have also tried swapping the template LED numbers. Now you have me wondering. That purple color when the device is off is odd to me. These devices connect so quickly to the wifi that I have not see anything that looks like a “searching for connection” color behavior that might indicate the colors are still not as intended.

akseidel:
I have tried all eight combinations for LED1 and LED2 on template parameters GPO01 and GPO13.
Using the convention DEVICE IS ON / DEVICE IS OFF, where Blue/Purple means BLUE IS ON / PURPLE IS OFF and the NONE means no color, we have:

Red/Purple : GPO01=Led2 (53), GPO13=Led1 (52)
Purple/Red : GPO01=Led2i (57), GPO13=Led1 (52)
Blue/None : GPO01=Led2i (57), GPO13=Led1i (56)
None/Blue : GPO01=Led2 (53), GPO13=Led1i (56)
Blue/Purple : GPO01=Led1 (52), GPO13=Led2 (53)
None/Red : GPO01=Led1i (56), GPO13=Led2 (53)
Red/None : GPO01=Led1i (56), GPO13=Led2i (57)
Purple/Blue : GPO01=Led1 (52), GPO13=Led2i (57)

Blue/None might have been the default firmware behavior but that setting does not provide a lighted button for manual use when lights are out if the device is used for lighting. I prefer the Blue/Purple setting.

Perhaps the red LED might be used by the default firmware to signal a communication condition. I do not see any color change behavior when the Tasmota firmware does not see a wifi signal.

PR #739, opened 2020-11-02
Update teepao_curtain_switch

Baldhor:
One year ago I proposed a better configuration. I can now confirm it's working much better than old one. So time has come to clean up 👍 
Also added backlog command.


blakadder:
Thanks for that!

PR #741, opened 2020-11-06
le_901001-US-a no longer compatible with Tasmota

shanecb:
The WB2L has an ARM MCU so these bulbs are no longer compatible with Tasmota. Here's the data sheet for the WB2L - https://developer.tuya.com/en/docs/iot/device-development/module/wifibt-dual-mode-module/wb2l-datasheet?id=K9duegc9bualu


blakadder:
thank you

PR #742, opened 2020-11-07
Adding mirabella_genio_I002940

maxious:
The model currently for purchase still supports tuya-convert and has full RGB light functionality available in Tasmota


blakadder:
thanks

PR #745, opened 2020-11-08
add maxico EDM-1WAA-EU Dimmer Switch

hensur:
adds maxico EDM-1WAA-EU. The device has a second controller chip for the dimmer, which is connected to the ESP via serial. It uses the TuyaMCU Protocol.


blakadder:
added from the form

PR #751, opened 2020-11-13
Update treatlife_DS01

BiOs2k:
Update information on new version of Tuya firmware that can't be flash with Tuya-Convert. only hard wire flash will work on those switch


blakadder:
thanks

PR #752, opened 2020-11-14
Update deta_6904HA

rabbit-aaron:
I think the status LED (led 5 on the PCB) is inverted.

I tried to assign it LED1, then LedPower1 state is inverted.
I had to assign it to LED1_i then LedPower1 state becomes correct.

So here I think we should assign GPIO0 to LedLink_i (158).
This might be the case for the other Deta wall switches too, but I only have a four gang one to test it on.

### DIFF
--- a/_templates/deta_6904HA
+++ b/_templates/deta_6904HA
@@ -8,7 +8,7 @@ flash: serial
 category: switch
 type: Switch
 standard: au
-template: '{"NAME":"Deta 4G Switch","GPIO":[157,0,0,19,18,21,0,0,23,20,22,24,17],"FLAG":0,"BASE":18}' 
+template: '{"NAME":"Deta 4G Switch","GPIO":[158,0,0,19,18,21,0,0,23,20,22,24,17],"FLAG":0,"BASE":18}' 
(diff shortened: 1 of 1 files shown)


rabbit-aaron:
There is a trick to these touch wall switches, if we use Button as the component for the touch sensors, the relay will only switch on/off after your figure leaves the sensors.

If we use `switch` instead of `button` and set `SwitchMode` to 3, it switches on/off when you put your finger on it.
I've also used this trick on the SONOFF TX2 wall switches, it is a much better experience.

On top of that, if you want to do something fancy, you can set SwitchMode to 5 to enable HOLD mode, I use the HOLD mode to control other lights in other circuits via MQTT using SwtichTopic + Rules.

for this particular model, this one command can change buttons -> switches and set switchmodes to 3.
```
BACKLOG SWITCHMODE1 3; SWITCHMODE2 3; SWITCHMODE3 3; SWITCHMODE4 3; TEMPLATE{"NAME":"Deta 6904HA","GPIO":[576,0,0,162,161,224,0,0,226,163,225,227,160,0],"FLAG":0,"BASE":18}
```

rabbit-aaron:
Should I add the above comment to the documentation?

PR #754, opened 2020-11-15
Update zemismart_YH002 template for tasmota 9.1.0

ArmoredBlood:
I manually flashed this device with the latest version of tasmota, so no idea if it would have worked with tuya-convert...
Regardless, I had to migrate the template to the new GPIO format to get it to work. I used the tasmota GPIO conversion page to do the pin conversions which worked great.

LMK if you'd like to see anything else modified


blakadder:
that will not work correctly

PR #757, opened 2020-11-18
Update fitop_SQPA6007-A

okh-mzny:
Got one of these bulbs today, wouldn't flash with tuya-convert, opened it up, found a WB3L.


blakadder:
thanks

PR #765, opened 2020-11-21
Add note about built-in module

Tediore:
There's no need to use a template for this device since a built-in module exists.


Tediore:
Thoughts?

blakadder:
none

PR #769, opened 2020-11-22
updated template to work with Tasmota 9.1

krikk:
made the required changes to make this template work under Tasmota >= 9.1


blakadder:
all old templates are converted automatically to new ones in tasmota

PR #770, opened 2020-11-23
Bump dates for plugs successfully flashed with Tuya Convert.

stevenewbs:
Bump the dates for some known good Tuya Convert UK plugs. Also some white space got trimmed.


blakadder:
thanks a lot

PR #773, opened 2020-11-24
Update bawoo_EUWL122130

ecadagiani:
The first template not worked at all for me, with Tasmota 9.10.
New template provided was on V2 version.
It seems to me, that this bulb is no longer for sale


blakadder:
new templates go into a new field `template9:`

ecadagiani:
Yes, it's done.

PR #777, opened 2020-11-29
Update Gosund WP6 template page.

netstx:


netstx:
@blakadder please let me know if you are ok with this update. thanks!

PR #781, opened 2020-11-30
Update ener-j_SHA5302

phedders:
Add alternate template - possible different version with alternate relay wiring. At least works correctly for an Ener-J Quve-330 that I recently bought from Amazon in the UK.
Using Tasmota 9.1.0

### DIFF
--- a/_templates/ener-j_SHA5302
+++ b/_templates/ener-j_SHA5302
@@ -14,6 +14,10 @@ standard: uk
 ---
 Left Hand Socket is Button 1
 
+There may be different versions where the relays are inverted and this template is required:
+```json
+{"NAME":"Ener-J 2-Gang ","GPIO":[33,0,0,0,0,224,32,0,225,320,0,0,0,0],"FLAG":0,"BASE":18}
+```
 There is a blue LED above the earth pin of each socket; these light when the corresponding socket is powered.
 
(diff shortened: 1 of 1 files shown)


blakadder:
do you perhaps have this version instead https://enerjsmart.co.uk/products/enerj-13a-smart-wifi-twin-wall-sockets-with-usb ?

phedders:
No! Its definately looks identical.
I have now bought some TCP TAUWIS2GWUK from Argos which are identical also, and have the same inverted switch arrangement. How should I add that device to your db? As a new device with link or as an alternative name?

https://www.argos.co.uk/product/7630222

Thanks for your DB!

phedders:
I would also like to add three warnings about this device!!!

1) The socket earth pins are connected to each earth screw lug and they are NOT cross-linked - you must link both earth screw lugs! There are copper "rings" exposed to the backing plate, but the plate is painted so they serve very little purpose unless you remove the paint and screw the plate in tight. Dodgy as heck.

2) On several of mine, the out1 copper contact was barely soldered to the board - when I removed the board on one the copper contact litterly fell out. I have gone around and resoldered them all!

3) I would rate the soldering/contacts/wire link as max 5 or 6A on these sockets.

So they are very useful - but "made to a price" - you are not getting MK quality here. Not even close to Contactum or BG.

PR #782, opened 2020-11-30
Update gosund_SW2

dmlau76:
Existing script would turn on the lights if the device was cold-booted due to a lengthy serial packet beginning with "72" upon initial boot.  The script was parsing this spurious packet and the "dimmer %f%" command was erroneously turning my light on every time the device rebooted.  The proposed changes check that the serial commands being acted on begin with "24", otherwise they are ignored.


blakadder:
thanks

PR #783, opened 2020-11-30
Update gosund_UP111

jerbzz:
Fix LEDs so that blue is power, red is status, and inverts are correctly set. Previously, purple = on, and red = off. Status was shown with a flash from purple to blue. Now blue = on, dark = off, and red = status. Also update to new GPIO numbering,


blakadder:
thanks

PR #784, opened 2020-12-01
Update C137

Cytomax55:
This is my first PR, i have no idea if i am doing this correctly but 
Tasmota 9.1 has breaking changes with the GPIO 
https://tasmota.github.io/docs/
Redesigned ESP8266 GPIO internal representation in line with ESP32 changing Template layout too
so you may consider having a template for 8.5.1 and before and 9.1 and after


blakadder:
tasmota converts gpio's if you're using an old template

PR #791, opened 2020-12-03
Update avatar_2314207011

dorbak:
Updated to add the 'W' component to the configuration (was only RGB before)


blakadder:
no, you updated it to be RGB+CCT

PR #793, opened 2020-12-05
Add flashing pinout (picture) for AWP08L

barbudor:
Provided by user on Discord
Could probably be added to all AWP08L variants (I cando that after if the PR is accepted)


blakadder:
no need, there are different AWP08L plugs

PR #811, opened 2020-12-14
Changed non-standard quote symbol

tinkermax:
The template contains "left quotation marks" instead of the expected normal quotes. This causes problems when trying to configure Tasmota, which could waste a lot of time for the unwary.


blakadder:
good catch! thank you

PR #812, opened 2020-12-14
Update orvibo_CT10W

BBBits:
correct template


blakadder:
that is not a correct template since its a v9 template in a classic template field

BBBits:
ok , can you modify please.
GPIO5 to IRrecv
GPIO14 to IRsend

PR #822, opened 2020-12-20
Update Wyze bulb

Quantum-cross:


Quantum-cross:
@blakadder I ALWAYS mix up the [] and () in markdown link syntax, can you fix that for me? :x

PR #844, opened 2021-01-01
Update frankever_FK-PW801US

sabatmonk:
this setting give expected results (led on when relay on, button toggle relay)


blakadder:
thank you

PR #853, opened 2021-01-04
Update avatar_AWP14H

chriscn:


blakadder:
which one is it then? new chipset or new tuya convert firmware?

chriscn:
The ones I bought at around 27/12/2020 had the ESP chipset but had the newest firmware.

PR #856, opened 2021-01-05
Add info about unsupported hardware

jnt2007:
I got a new hardware version with unsupported wifi chip. Just sharing information.


blakadder:
thanks for the info

PR #857, opened 2021-01-05
Update mixigoo_10W_E27

curlyel:
The device is actually a RGB+CCT bulb. The proposed changed template is confirmed on 9.2.0 (lite).

### DIFF
--- a/_templates/mixigoo_10W_E27
+++ b/_templates/mixigoo_10W_E27
@@ -6,7 +6,7 @@ category: bulb
 standard: e27
 link: https://www.amazon.de/gp/product/B07VZXDX6J
 image: https://user-images.githubusercontent.com/5904370/66717700-49e76400-eddc-11e9-87f4-2e3b2667eac0.png
-template: '{"NAME":"Mixigoo Bulb","GPIO":[0,0,0,0,37,40,0,0,38,0,39,0,0],"FLAG":0,"BASE":18}' 
+template: '{"NAME":"Mixigoo Bulb RGBCCT","GPIO":[0,0,0,0,416,419,0,0,417,420,418,0,0,0],"FLAG":0,"BASE":18}' 
 link2: 
 flash: tuya-convert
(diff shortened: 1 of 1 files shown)


blakadder:
how did you come to that realisation, the item description clearly states 6500k+RGB?

curlyel:
You are right, even the print on the socket indicates that it is RGB plus "CW":



But I found a template here: https://github.com/stefanbode/Sonoff-Tasmota/blob/master/TEMPLATES.md

`> Mixigoo 950lm            {"NAME":"Mixigoo Bulb","GPIO":[0,0,0,0,41,38,0,0,39,0,37,40,0],"FLAG":0,"BASE":18}`

indicating five GPIO's used to control the bulb. Unfortunately, this template was not working correct ...
But I remembered the template on your site and moved back to it:

`{"NAME":"Mixigoo Bulb","GPIO":[0,0,0,0,37,40,0,0,38,0,39,0,0],"FLAG":0,"BASE":18} `

This was working o.k. for the RGB part _and_ cold white. So I left PWM1/2/3/4 and started experimenting with PWM5 (and SetOption92 ON/OFF) to find out if some other GPIO is used and if PWM-CT is deployed as well. Finally I figured out, that GPIO13 simply controls warm white without PWM-CT.
_Btw.: The bulb can make a quite nice warm white light... ;-)_

image

The bulb where bought following the Amazon-link here https://templates.blakadder.com/mixigoo_10W_E27.html
 
When you compare the pictures, you may notice a slight difference on the marking:

image
image

So probably, there is a new hardware revision sold with the same name...
Maybe, you could add them as "Mixigoo 10W RGB-CCT" in addition the old (?) version?

Here is short video showing the transition from cold to warm white (sorry for the poor quality ;-) )

https://user-images.githubusercontent.com/11849305/103701668-b5e85600-4fa6-11eb-8e58-f1ce8ea5d799.mp4

PR #866, opened 2021-01-08
Added RGB version of Wipro smart Batten

rishabmehta7:
Added RGB version of Wipro smart Batten. hence renames the white one with a -01 suffix and also specified 'white' and 'RGB' in title.


blakadder:
template for rgb isn't correct

rishabmehta7:
Any way to figure it out without opening up the device?

blakadder:
that's a question for Tasmota support. I'm more interested in why would you add a template that is wrong?

PR #872, opened 2021-01-11
Update holman_CLXRGB60

co2au:
added in push button (GPIO0) and template8


blakadder:
and what is template8?

co2au:
Sorry, just getting used to github and the templates. My bad. Shoult have been "template" not "template8". Guess you worked that out?

PR #875, opened 2021-01-11
Update holman_CLXRGB60

co2au:
Template for tasmota 8.x.x with pushbutton on GPIO0


blakadder:
that template isn't displayed anyway

PR #884, opened 2021-01-16
Update gosund_SL2

pnposch:
just successfully flashed with tuya-convert, a lot easier than my soldering skills


prodigy7:
Thank you for supporting the gosund SL2 light stripe. Can you describe how I can get the light stripe into the flash mode? Try it with pressing the button on the remote, the stripe is flashing for some seconds but tuya convert wouldn't find the light stripe.
Light stripe is new, still not connected with app and no firmware update was done.
Looking into smarthack-psa.log shows messages which could indicate, that my stripe has always a newer firmware (see https://github.com/ct-Open-Source/tuya-convert/wiki/Collaboration-document-for-PSK-Identity-02). If there is a way with newer firmware, could you describe at https://github.com/blakadder/templates/blob/master/_templates/gosund_SL2 how?

Thank you :-)

PR #891, opened 2021-01-21
Converted values to new GPIO format

daj102:


blakadder:
new format goes under template9: field and the old format is kept

daj102:
Hopefully this is what you meant, @blakadder ??? Sorry if I missed something about the GPIO template updates - was not aware of the template9 field.
--DJ the n00b :)

blakadder:
great, thanks!

PR #892, opened 2021-01-21
Fixed name to align with MJ-S01, and updated to new GPIO format

daj102:


daj102:
Hopefully this is what you meant, @blakadder ??? Sorry if I missed something about the GPIO template updates - was not aware of the template9 field.
--DJ the n00b :)

PR #903, opened 2021-01-27
Update lilygo_T-Internet-POE

yoyanec:
EthClockMode 3 is correct mode for this board. EthClockMode 1 is not working.


blakadder:
thanks

PR #904, opened 2021-01-28
Update geeni-GN-BW911-999

cantfocus:
The default PwmFrequency of 977 was causing my bulbs flickering when dimmed.  Adding a note for how to change that.


blakadder:
thanks

PR #908, opened 2021-01-29
Description and pictures added

TDCroPower:


TDCroPower:
The information about the Blitzwolf LT11 RGBW Strip was not much so I added my experience.
I hope my pull request is done correctly.

Where can you actually add pictures that you use for the explanations?

TDCroPower:
@blakadder thanks for the merge!
do you know why the added images are not displayed?

blakadder:
they're missing "?raw=true" after .png

PR #911, opened 2021-01-31
Update yeelight_YLDL01YL

elajoie:
added current workaround


mrdvt92 (changes requested):
Please format all dates in ISO format e.g. 2021-01-31

PR #913, opened 2021-02-01
Update arlec_ALD233HA

Final-Hawk:
Added warning that this device starts coming with a tasmota incompatible chip.


Final-Hawk:
No rush, but when will/if this be merged?

blakadder:
why ask if there's no rush?

Final-Hawk:
> why ask if there's no rush?

Fair point haha

PR #918, opened 2021-02-03
Wyze OTA support.

elahd:
A bit of self-promotion. I created a tool to flash the Wyze Bulb and Wyze Plug OTA. This PR adds a note to the pages for those devices with a link to the repo.

As an aside, this tool (theoretically) can also be used for other ESP8266 devices that use the ESP8266_RTOS_SDK with the v3 bootloader / OTA workflow.


blakadder:
Cool project, always nice to see easy flashing methods!!!

PR #922, opened 2021-02-07
Add notes about changing boot state

JulioSDDR:
Due to quirks with this device, it is not immediately clear how to automatically turn on only the light on boot up. This workaround was suggested by ascillato2 at Tasmota issue #6193


papapep514:
I am getting "ON" state when its  the state is OFF".

PR #936, opened 2021-02-27
Update martin-jerry-MJ-S01

DentonGentry:
MJ-S01 purchased 24 Feb 2021 from amazon.com no longer flashable using tuya-convert.


DentonGentry:
Nevermind, after three tries it did work.

PR #947, opened 2021-03-05
Update ESP01

kruebsam:
1.) Pin used to drive WS2812 is GPIO2 of function type 1376
2.) the link to the ESP-01 and ESP-01S difference does not at all mention the ESP-01 but the ESP-07. That´s why I did remove that link


blakadder:
what does WS2812 have to do with ESP01?

kruebsam:
Oh yes, regarding ws2812 you are so true. Sorry for that one. OK, get me your paypal-donation link and I´ll try to place a small compensation for my mistake.

PR #949, opened 2021-03-05
edit template gosund_SP112

matt8707:
previous template is not working
tested with tasmota 9.3.1

https://github.com/arendst/Tasmota/blob/development/TEMPLATES.md


blakadder:
which hardware revision is that?

matt8707:
I have no idea..

PR #951, opened 2021-03-06
Update Digoo sg-sp01

fhloston:
Correct PWM mapping, setoption37 96 not necessary.
New 16-bit GPIO Numbering scheme.


blakadder:
thanks

PR #956, opened 2021-03-07
Update WL_SW01_16.md

PJO2:
You mention this device, a sonoff TH-16 clone, as unsupported.  Hopefully they can be upgraded to Tasmota with a bit of patience ! 
Not sure to do it right, but i create a pull request if you want to update the status.
Anyaway thanks so much for Tasmota !


blakadder:
Did you order the device from the same link?

PJO2:
I ordered them from the aliexpress Coolego strore (https://www.aliexpress.com/store/2187130). But they have retired the product. However you can find some units by searching for "TH-16" (https://www.aliexpress.com/wholesale?catId=0&SearchText=TH-16)

For these switch, they are selling APIs, but, to make it work, you have to create an account on a chinese cloud server and every API has to by authenticated by this chinese server. I was very happy to be able to flash Tasmota !!

PR #959, opened 2021-03-14
Update mirabella_genio_I002741

solyxius:
Recently purchased device was unable to flash wirelessly only via serial. 

PS - Only alternative template worked, primary template had no light output from the device whatsoever.


blakadder:
thanks

PR #962, opened 2021-03-16
Update Dogain 320lm

JVital2013:
New bulbs contain a BK7231TQN32, as purchased on Amazon in March 2021. Not compatible with Tasmota as far as I can tell.  I took one apart - if anyone wants images let me know


blakadder:
thanks for reporting

PR #963, opened 2021-03-17
Fix up confusion for Arlec 9W LED Downlights

talltechdude:
The existing template was marked for model numbers and links for the CCT version but had a template generating an RGB+CCT configuration which is actually for a light an Arlec `ALD092RHA`.

Split out the two as separate products to clear up any further confusion.


blakadder:
thanks

PR #966, opened 2021-03-18
tuya-convert didnt work for me

LocXar:
Did try to Flash it with the newest Tuya-Convert but got the Message, that the Firmware is to new. If someone can show me how to proof it, i will proof it for you :)


LocXar:
Its my first pull request on GH. If i did something wrong, explain it to me and i try to do it right.

blakadder:
thanks

PR #970, opened 2021-03-22
Added a link to a flashing jib

dimaj:
and added a small correction to the GRND, TX and RX pin location on the chip.


dimaj:
that was fast!
Thanks!

PR #978, opened 2021-03-28
Add Description for pins

jasonyork:
Took me a while to track down the order of the test points / pins.  I did find this image here:
https://github.com/arendst/Tasmota/issues/2789#issuecomment-415563022
However, found that TX and RX were reversed (eg, TX on the Shelly should be RX on the programmer.)  

This PR adds the correct order of the pins to this template to hopefully help others.


blakadder:
thanks for the contribution

PR #982, opened 2021-04-01
Update goldair_sleepsmart_GCPF315

Final-Hawk:
This Pull request adds  percentage speed control for home assistant core-2021.4.0 as the old way using speed is deprecated and due to be removed in later versions


blakadder:
thanks

Final-Hawk:
All good, took me awhile to figure out so I thought I would update it here!

PR #983, opened 2021-04-01
Update arlec_ALD233HA

Final-Hawk:
Updates arlec light strip with instructions for when you change the chip.
Hopefully is correct and elaborate enough, if not please suggest changes!


blakadder:
new template goes in under `template9:` field

Final-Hawk:
Like that?

Final-Hawk:
The page seems to have disappeared. Are you seeing the page on your end?

PR #984, opened 2021-04-02
Update diolamp_A609WIFI

costakis79:
Changed the SetOption37 value to 30 because the Ww,Wc order was inversed.


blakadder:
ok thanks

PR #986, opened 2021-04-03
Updates to DETA 6914HA Fan Speed Controller with Light Switch

forfuncsake:
- Invert the LinkLed so it's not always on
- Optimise the rule to reduce delay turning relays 3 and 4 on after relay 2 when using the button or events.

Note that to add the additional event for fan power toggle, I needed to reorder the rule lines, use lower case and shorten the event names to fit everything in a single rule.
Interestingly, the same rule set in a different line order or with upper case causes the Tasmota unishox compression to be less efficient.

I found that with 5 minute telemetry, the power state in HA can be a bit laggy, but sending power and speed commands works as expected. The rule is flexible enough to handle someone sending commands to POWER2 instead of sending events, but the relay 3 and 4 response will be slightly delayed as it is before this change.

### DIFF
--- a/_templates/deta_6914HA
+++ b/_templates/deta_6914HA
@@ -102,8 +102,8 @@ Rule Logic
   payload_low_speed: "Event s=1"
   payload_medium_speed: "Event s=2"
   payload_high_speed: "Event s=3"
-  payload_off: 'OFF'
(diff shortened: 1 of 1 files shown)


blakadder:
compression works better since different caps take more room because they're different

PR #988, opened 2021-04-04
Update sonoff_4CH_Pro

LexxM3:
Identifying a compatible clone and its basic conversation details.


blakadder:
clone is not a sonoff 4ch anymore, is it? add it as a new device

PR #993, opened 2021-04-07
Correction for athom_PG04-UK16A template

ralphrmartin:
Existing template is wrong. Please see manufacturer's page here
https://www.athom.tech/blank-1/uk-plug

The template given causes the relay to switch on and off at 1 second intervals!


blakadder:
thank you for the correction

PR #995, opened 2021-04-08
Update tcp_WISSINWUK

bmerrihort:
Add some information about a different version of the TCP Smart plugs that is on the market.


blakadder:
good information to have, thank you

PR #998, opened 2021-04-12
GD-30W: update Home Assistant fan to be consistent with new percentage speeds

Nardol:
Since Home Assistant 2021.3, fans nammed speeds are deprecated.
Speeds are now specified by percentage, with backward compatibility until version 2021.7.


blakadder:
thanks

PR #1004, opened 2021-04-16
GD-30W: update RGB rule and water sensor

Nardol:
Rule1 is updated to make RGB cycle switch work again.
Rule2 allows making Power4 working when the diffuser is empty.
All credits go to the international Home Assistant community forum members who have provided all these updated rules whichI tested successfully.


Nardol:
As I have no feedback for this PR, I taken the oportunity to update water sensor code to be consistent with last update from @SirGoodenough and I added working code when RGB cycle/fixed color has been activated.
I tested it and it works right.

I'm going to open another PR for Maxcio 400ML diffuser to document these codes.

@blakadder please tell me if I made something wrong :slightly_smiling_face:

Nardol:
I've merged every commits into one in the case it would be better to check this.
All tests pass for me.

blakadder:
That setup is terribly outdated, I added new configuration options

PR #1021, opened 2021-04-23
Update ESP12F_Relay_X4

phedders:
Add note to avoid IO16 if required.

### DIFF
--- a/_templates/ESP12F_Relay_X4
+++ b/_templates/ESP12F_Relay_X4
@@ -14,4 +14,4 @@ standard: global
 ---
 Headers must be soldered and jumpers needs to be installed for enabling relays. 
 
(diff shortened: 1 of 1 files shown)


tjhiwa:
I tried the template but it doesn't work.
What do you mean by "Headers must be soldered and jumpers needs to be installed for enabling relays."
Could it work just with flashing of Tasmota?

phedders:
On Wed, 2021-07-07 at 06:26 -0700, tjhiwa wrote:
> I tried the template but it doesn't work.
> What do you mean by "Headers must be soldered and jumpers needs to be
> installed for enabling relays."
> Could it work just with flashing of Tasmota?
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.

The board works - but the relays will not...

As supplied they are not connected to any GPIO pins on the ESP. The
headers allow you to link the four relay trigger pins/pads to four
GPIOs. The easiest way is to put the 2x4 row of headers on in the
middle and put four links on.

This does have the GPIO16 problem - but it may not be a problem for
you, and if it is, you only need to add a link wire from a different
GPIO to that relay trigger pin/pad.

PR #1023, opened 2021-04-24
Added Arlec MAL315HA security light

Lenbok:
(From the outside, looks the same as MAL300, re-using the image)

Here's the main board:
IMG_20210413_205351

I don't yet have an image showing the post-transplant.

### DIFF
--- a/_templates/arlec_MAL315HA
+++ b/_templates/arlec_MAL315HA
@@ -1,0 +1,78 @@
+---
+date_added: 2021-04-18
+title: Arlec Smart Security Light
+model: MAL314HA
+image: /assets/images/arlec_MAL300H.jpg
+template: '{"NAME":"Arlec MAL315HA","GPIO":[0,107,0,108,0,0,0,0,0,0,0,0,0],"FLAG":0,"BASE":54}' 
+link: https://www.bunnings.com.au/arlec-20w-grid-connect-smart-security-light-with-movement-activated-sensor_p0223575
+link2: https://www.bunnings.co.nz/arlec-20w-grid-connect-smart-security-light-with-movement-activated-sensor_p0223575
+mlink: 
+flash: Tuya WB3S
+category: light
+type: Light
+standard: au
+---
+
+## Features
+
+* Light duration: How long in seconds the light stays on after motion is detected. From user manual: min 10 sec to max 15 min.
+
+* Sensitivity adjustment (amount of movement required to trigger): Min = most sensitive, max = only detect large motion.  To tune, start with at minimum, and have someone move while increasing sensitivity.
+
+* Light level (lux adjustment): Min = day or night, max = night only.  For normal operation, adjust appropriately (e.g.: wait until the time of day, set sensitivity to most sensitive, lux to max, and while someone walks around reduce lux threshold until it triggers).
+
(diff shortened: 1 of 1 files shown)


Lenbok:
In the description I list a rule that I thought would permit the Toggle button in Tasmota web UI to switch the light between manual (on) and auto (usually-off), however it doesn't quite work -- If the light is in auto mode and the sensor activates, the act of it turning the light on causes the rule to fire, swapping the light out of auto and into manual mode (thus the light stays on).  Can someone suggest how to fix that behaviour?

Lenbok:
For future reference, here's a picture after replacing the W3BS with an ESP-12 and soldering appropriate jumpers.

IMG_20210504_200655

david-collett:
> In the description I list a rule that I thought would permit the Toggle button in Tasmota web UI to switch the light between manual (on) and auto (usually-off), however it doesn't quite work -- If the light is in auto mode and the sensor activates, the act of it turning the light on causes the rule to fire, swapping the light out of auto and into manual mode (thus the light stays on). Can someone suggest how to fix that behaviour?

I think I've figured this out with the following rules:
```
Rule1 ON TuyaReceived#Data=55AA03070005660100010177 DO Rule2 0 ENDON ON TuyaReceived#Data=55AA0307000565040001027A DO Rule2 1 ENDON ON TuyaReceived#Data=55AA03070005660100010076 DO Rule2 1 ENDON

Rule2 ON Power1#State=0 DO TuyaSend4 101,1 ENDON ON Power1#State=1 DO TuyaSend4 101,2 ENDON
```

The idea is that when ..77 is received (relay ON) the rule is toggle rule is disabled to prevent it firing on the power state event, but if ..7A is subsequently received (enum->2) indicating it was turned on with a toggle, it is reenabled (to allow toggle to turn it off again). Then it is in auto mode and the light is triggered by the sensor, 77 is received but *not* 7A, so Rule 2 remains disabled until the light times-out and 76 is received which re-enables the rule. Seems to be working for me!

Lenbok:
@david-collett Awesome, I've been trying your new rules out and they seem to be working great! 

I haven't attempted it yet, but I'm guessing that a modified Rule2:
```
Rule2 ON Power1#State=0 DO Backlog TuyaSend4 101,1; TuyaSend1 106,1 ENDON ON Power1#State=1 DO TuyaSend4 101,2 ENDON
```
should make it so that after manually toggling to ON it would fire the one hour auto timer. I'm guessing that part would work but we may need another clause in the first rule so that Tasmota would get notified when the unit automatically switches back from manual to auto?

david-collett:
Hmm, I don't think Rule1 needs to change because Rule 2 will already be enabled as it should be at that point.
Rule 1 only needs to disable Rule 2 when the relay is triggered by an motion event, and then only until the light goes out again.
I haven't installed my light yet (hopefully tomorrow), then I can test it out a bit more.

Lenbok:
@david-collett Indeed, seems to work from my testing so far. Have opened #1055 to update the description text.

david-collett:
It appears that any custom settings (e.g, duration, sensitivity) are lost after a power cycle. So I've proposed a new rule to restore them on boot in #1061

PR #1034, opened 2021-05-02
Update shelly_Dimmer_2

Bascht74:
Add some hints from https://github.com/jamesturton/shelly-dimmer-stm32/issues/11 and https://github.com/arendst/Tasmota/issues/6914


blakadder:
thanks

PR #1035, opened 2021-05-03
Update sinilink_XY-WF36V

flywire:
Add image of bottom of module showing pin holes.

Need to verify this displays correctly.


flywire:
I'm not clear exactly how you get a serial connection to this module and if Win10 requires a driver to access it. Is it accessible by wifi from factory?

flywire:
How about adding a link to that blog post then?

flywire:
XY-WF5V

See https://templates.blakadder.com/sinilink_XY-WFUSB.html for flashing notes.

PR #1036, opened 2021-05-03
Update treatlife_DS03 for new HASS MQTT Fan syntax

boybert:
Update the HASS Fan YAML for new syntax (no more speed payloads, using preset modes instead as illustrated here: https://community.home-assistant.io/t/mqtt-fan-speed-issue/34259/6)


blakadder:
excellent, thanks a lot for that fix!

boybert:
Happy to help! Thank you for the excellent resource!

PR #1040, opened 2021-05-07
Update geeni-GN-WW104-199

heegard:
The existing template-alt doesn't control the LED correctly.  The correction makes the LED follow the state of the switch.


heegard:
I want to withdraw this change.  The template is for the model GN-WW108 which has the same look and name, but has a different model number.

PR #1058, opened 2021-05-23
Correct model number typo

Lenbok:
It's actually MAL315HA, and is correct in most places but there is a typo in the model number metadata section.


blakadder:
Thanks for the fix

PR #1060, opened 2021-05-24
Update hbn_U151T with incompatible module

ckkeane:
U151T plugs are now shipping with WB2S modules which cannot be updated with Tasmota. Added a warning to the page description, and updated with my purchase date and location.


blakadder:
thanks

PR #1061, opened 2021-05-25
Update arlec_MAL315HA

david-collett:
Add note and example for restoring MCU settings on boot (Duration, Light sensitivity etc).


Lenbok:
Nice!

PR #1066, opened 2021-05-29
add button, fix led

mvadu:
Previous template did not handle the button. Adding button. Also the led was inverted. Fixed.


blakadder:
thanks

PR #1071, opened 2021-06-01
AOFO C730: Add tuya-convert notice

rmelotte:
I bought this device on amazon.de just a few days ago, and it turns out tuya-convert doesn't support the revision I have yet:
```
Your device's firmware is too new.                                                                                                                                                                    
Tuya patched the PSK vulnerability that we use to establish a connection.                                                                                                                             
You might still be able to flash this device over serial.                                                                                                                                             
For more information and to follow progress on solving this issue see:                                                                                                                                
https://github.com/ct-Open-Source/tuya-convert/wiki/Collaboration-document-for-PSK-Identity-02
```

While at it, also add the model to the template.

### DIFF
--- a/_templates/aofo_4AC4USB
+++ b/_templates/aofo_4AC4USB
@@ -1,6 +1,7 @@
 ---
 date_added: 2021-03-14
 title: AOFO 4AC 4USB
+model: C730
 category: plug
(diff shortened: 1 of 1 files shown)


blakadder:
thanks

PR #1074, opened 2021-06-07
fix LED

mvadu:
Found problems with LED config. Looking at pcb traces, GPIO2 is not connected to anything, its missing a via, and GPIO0 is controlling both LEDs. So have to set the GPIO2 to none so firmware can actually show error conditions accurately.


mvadu:
WIN_20210607_12_39_26_Pro
WIN_20210607_12_39_26_Pro

PR #1078, opened 2021-06-09
Update H801 G<->B on IO12 & IO13

jyoci:
I have found that blue and green colors where interchanged into the H801 with respect to the label on the top of the connectors.
Green = PWM 2 = IO12
Blue   = PWM 3 = IO13


blakadder:
the template is based on module 20 configuration which is as is since it was introduced in Tasmota

PR #1092, opened 2021-07-02
Corrected typos

jceccato:
Corrected model number and name in template (From: CSH-E14WW10W to CSH-E14WW5W)
Added alternative sister model.
Removed lumens from title to match other Connect SmartHome Products


blakadder:
thanks

PR #1118, opened 2021-08-15
Update luci-connect-remote-control

girkers:
There was a . instead of a , which rendered the LOW functionality NON functional. Line 47 relates


blakadder:
thanks

PR #1120, opened 2021-08-16
Added internal RST button to Shelly i3

Olifant1990:
Added internal RST button to Shelly i3

Not tested in Tasmota, found in own firmware, is component 17 the right GPIO config for GPIO 4?
Component says "Button active low, internal pull-up resistor"
The button needs to be pulled-up and 0 is pressed, 1 is not pressed.


blakadder:
i've added it this way since it is untested

PR #1126, opened 2021-08-21
Add secondary link for Gosund UP111

HandyHat:
For some reason the first link does not appear at https://templates.blakadder.com/gosund_UP111.html


blakadder:
thanks

PR #1142, opened 2021-09-04
Better flashing link for Elehot AWP16L

sylvinus:
I propose replacing the generic archive.org link which was not very useful for me with a step-by-step walkthrough and pinout for this specific device.
Thanks!


blakadder:
nice article

PR #1143, opened 2021-09-05
FIX - replace corrupted image 

robertoostenveld:
for https://templates.blakadder.com/mirabella_genio_I002577.html

- it is unclear why the image is corrupt, but it applies both to the online version and the one I checked out from github
- I retrieved the image from the cache of google image search


robertoostenveld:
I noticed that the image for https://templates.blakadder.com/ZN281402.html was also corrupt, I did the same for that one.

blakadder:
thanks

PR #1144, opened 2021-09-06
Update sinotimer_TM608

ouinouin:
Hi , 
got 6 of them all with WB3S module. i dont know if the complete swap is possible , we should compare with an older tM608 with esp8266 to sse if they reworked the wifi daughterboards.


blakadder:
That's a shame. WB3S can be replaced with an ESP-12 type module. ESP-12S being the best choice due to having strapping pins already pulled correctly.

ouinouin:
Hi, I ll buy a bunch of esp12 /12S  and give soldering a try ( not easy to desolder with only a soldering iron.)
i ll update template if successful.
I just de-soldered the daugherboard containing the BL0937 , a think there wont be any issues.

PR #1145, opened 2021-09-06
Update for Youngzuth SW03-02 with WB3S

zeroping:
I got an Youngzuth 3in1 Switch, SW03-02, and it has a WB3S in it. As I understand it, it's unsupportable, but can be physically swapped with an ESP12. 

The first commit updates the Youngzuth directly, the second updates two very similar-looking Sana devices that may be affected the same way.

Hopefully the format of the edits look OK.


blakadder:
thanks, well done!

PR #1146, opened 2021-09-09
Update luminea_ZX-2858

fschaeck:
Changed the template to actually include the power meter and use the right GPIO for the button (GPIO0 and not GPIO3 as it was before)


blakadder:
thanks

PR #1147, opened 2021-09-10
Update Asakuki instructions

danimaciasperea:
Updated flash procedure.
Included home assistant configuration for light with effects.


blakadder:
thanks

PR #1153, opened 2021-09-19
Update gosund_UP111

danstreeter:
Added update to succesfull flashing of Gosund UP111 from recent purchase


blakadder:
thanks for the info

PR #1161, opened 2021-09-25
Update treatlife_DP10

AlbertoFedeli:
New board version is easier to flash without needing to remove the board or solder.


blakadder:
thanks

PR #1169, opened 2021-10-03
feat(Sygonix-SY-4276902): flashing instructions and pinouts

hertzg:
Includes the following information:
* Disassembly and flashing instructions
* Annotated Photos of the device & Pinouts


blakadder:
thanks

PR #1171, opened 2021-10-06
Update nedis_WIFIDS10WT

ignisf:
This commit expands the WIFIDS10WT documentation with instructions how to track the TEST, MUTE and LOW BATTERY messages.

It also gives instructions on how to integrate with Home Assistant


blakadder:
thanks

PR #1172, opened 2021-10-08
Update gosund_SW6

hi-flyer:
Original Rule2 was not accepted  due to extra Backlog misspelled Powe#state and : rather than ;.


simon-chadwick:
Many thanks - this was driving me crazy! FYI, I added another pull request, for setting the LED red when the circuit is open.

PR #1182, opened 2021-10-12
added instruction to open and some images of the inside

jnsw:


jnsw:
@blakadder did you manage to flash it or what is the source of that image you added in the first place?

PR #1206, opened 2021-11-01
Novostella UT55509 Template Update

fivesixzero:
Another quick template update based on my experience with a UT55509 yesterday.

That board-to-board interconnect is one of the most awful designs I've seen in a bulb so far, oof. 😂 

Not just saying that because I totally trashed the RGB feature on one of mine, of cousre. 🙄


blakadder:
good info ,thanks

PR #1208, opened 2021-11-05
Update lilygo_T-Internet-POE

kymodoke:
Boards version 1.0 and version 1.2 have different Ethernet ClockMode.
_EthClockMode 3_, for board version 1.0 and _EthClockMode 1_, for board version 1.2


blakadder:
I recommend setting separate template pages with notes on how to recognize the hardware revisions since LilyGO likes to revise their hardware a lot without changing the product names

kymodoke:
Hardware revision is clearly labeled on the board's PCB (on the back side).
examples v1.2 vs v1.0: https://imgur.com/a/KFv9LQu
and as an easy reminder, there is a cartoon cow on the back side of v1.2 ;)

blakadder:
ah ok then, thanks for the additional info

PR #1210, opened 2021-11-06
Update le_904102-RGB

DaKarlii:
Add some Information to use the built in Hardwarebuttons of this module


blakadder:
great addition, thanks!

PR #1211, opened 2021-11-06
Update supernight_dual

tony-fav:
updated template to include second button function and added my TC success date.


blakadder:
thanks for the fix

PR #1223, opened 2021-11-12
Update lilygo_T5-4_7

LUC4R4T0R:
Removed komma, which prevented the epdiy library from beeing used in compilation.


blakadder:
thanks

PR #1232, opened 2021-11-21
Update topgreener_TGWF15RM

ams2121:
Added update on ability to flash with tuya convert and level of effort if attempting serial flashing


blakadder:
great info, thanks

PR #1236, opened 2021-11-23
Update bn-link_BNC-60_U133TJ

racordoba22:
I had the same problem with the W2BS chip and I bought the WT-01N chip, which is an exact replacement, I programmed it with tasmota and installed it, it worked perfectly


blakadder:
thanks for the info!

PR #1238, opened 2021-11-23
Update gosund SW9

kruzer:
Modified template to control leds, report tuya-convert is working on this swich


blakadder:
this is a weird and very personalised use of LED's

PR #1252, opened 2021-11-29
Show Link Status with Blue LED

jscott304:
GPIO13 controls the Blue LED, The Red LED is driven in parallel with the Relay.


blakadder:
thank you for your contribution

PR #1253, opened 2021-11-30
Update energizer_EOX3-1001-BLK

crossan007:
I purchased a few more of these devices since adding the template, and discovered the steps I previously submitted can be simplified / better explained.


blakadder:
thanks a lot

PR #1258, opened 2021-12-01
Update Honeyzy DN20 with circuit board images

cizzop:
Update Honeyzy DN20 with circuit board images and some board information


blakadder:
thanks

PR #1261, opened 2021-12-03
Update rjwf-02a

sfromis:
Better include required config details in the template

### DIFF
--- a/_templates/rjwf-02a
+++ b/_templates/rjwf-02a
@@ -2,7 +2,7 @@
 date_added: 2019-12-06
 title: RJWF-02A
 image: https://user-images.githubusercontent.com/5904370/71354140-4cec9680-257b-11ea-8f6f-fad7d6c5858e.png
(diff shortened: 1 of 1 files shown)


blakadder:
Does that work on old templates?

sfromis:
Looks like the answer is yes. On current dev Tasmota, I did a test with the amended template on that dimmer, after setting DimmerRange differently. The console showed a message about the template being converted, and after the restart, the DimmerRange was reset according to the template.

sfromis:
More generally, when fixing small irregularities i templates, I've assumed that it is not preferred to convert them to new component numbers. While it would be easy to do, it does reduce compatibility with old binaries. While probably not much of a problem, my stance became to not change more than I came for. I'm open to suggestions about changing this stance :-)

blakadder:
That's fine, just wanted to check whether the cmnd key works even if the old template format is used

PR #1293, opened 2021-12-28
Update teckin_SB60

greeeaaasy:
HI,
i have an addition to the Teckin SB60. Hardware replacing with ESP-01D (8285) needs different Template. Worked for me.


blakadder:
why are you replacing the wifi module?

greeeaaasy:
> 

Because they cnahged the original Module to an "unflashable".  At least my tries showed no efford.

blakadder:
which is inside originally? any pictures?

greeeaaasy:
Yes, i found a WBLC5 - which i replaced.
teckin_sb60_orig
teckin_sb60_orig2

PR #1296, opened 2021-12-30
Wrong button mode gosund SL2

marc577:
Fixed wrong button mode. Inverted button leads to full reset after 40 seconds.
Newer models of LED strip seem to use other button.


blakadder:
thanks

PR #1299, opened 2022-01-03
Fix Shelly EM Template

barbudor:
Apply the one listed in https://github.com/arendst/mgos-to-tasmota

The one registered is using new GPIO numbering which then get badly translated


blakadder:
merci

PR #1301, opened 2022-01-05
Update stitch_35753

Taigitsune:
Changed base template to YTF IR Bridge (62) which disables serial in order to prevent a flood of stat messages in MQTT.


blakadder:
Thanks

PR #1304, opened 2022-01-10
Removed wrong PWMCT from Yeelight Meteorite

sfromis:
Removed SetOption92/PWMCT, as the lamp has 5 plain PWM channels, not with white split into brightness and CT. The misleading config made no difference for a long time, but since a month ago, Tasmota actually use the PWMCT setting, making the lamp misbehave on the white part. Also tweaked the text a bit for better info.


sfromis (commented):
Oops, `SetOption37 128` (PWMCT) should be `SetOption37 128` (ChannelRemap)

PR #1306, opened 2022-01-10
Update yeelight_YLDL01YL to save energy

sfromis:
After finding out how to save >1W of power while the lamp is idle, I automated it :-)


blakadder:
cool

PR #1309, opened 2022-01-13
Update ESP-01S-Relay-v4

BlazejosP:
New template now D4 GPIO2 is visible as user configurational and user can chose what he d'like to add there


blakadder:
Is a GPIO2 pin available anywhere on the board?

BlazejosP:
Yes but not straight directly I solder pin to socket for ESP2866 from underneath

PR #1311, opened 2022-01-18
Update treatlife_DS03

apetrycki:
Added some more information and options.  Corrected Rule1 so that it takes the user's configured topic.


blakadder:
Thanks

PR #1314, opened 2022-01-19
Update athom_CB01-TAS

tarontop:
remove switch 1


blakadder:
why?

tarontop:
> why?



> why?

It's the same as Sonoff Basic, "switch 1" is redundant because it has "Button 1"

PR #1327, opened 2022-01-27
Update zemismart_KS-811_1gang

timwelchnz:
Advise new purchasers that new versions are totally incompatible with Tasmota as they no longer contain an ESP8266 chipset


blakadder:
thank

PR #1331, opened 2022-01-29
Update arlec_GLD112HA

andybotting:
Update the Arlec GLD112HA template as PWM2 needs to be inverted for warm/cold to be mapped correctly, and add some details in the documentation.


blakadder:
Great addition, thanks a lot

andybotting:
Thanks, but all credit to you for hosting this valuable resource.

There's a few more very helpful bits of info I'm going to add shortly - so expect them soon.

cheers!

PR #1332, opened 2022-01-29
device no more available via links proposed

stilsch:
device no more available via links proposed


blakadder:
and what if it gets restocked?

PR #1360, opened 2022-02-24
New hardware version found with WB2S

OlliKantola:
New version with WB2S chip found. Direct replacement TYWE2S  was used and button pin was found to be at GPIO03. Energy metering chip identified as BL0937 and was able to accurately calibrate power monitoring when using BL0937 CF (2720) -option.


blakadder:
you should move the template for the replacement below your description of the new chip and leave the original template

PR #1382, opened 2022-03-22
Update martin_jerry_US-SD-KN01

MidnightLink:
Martin Jerry has official Tasmota flashing instructions for this device depending on what version of PCB. Added their PDF linked on their website (may need to be imported into assets)


blakadder:
thanks a lot for the addition

PR #1384, opened 2022-03-27
Update ce_smart_home_LTS-6A-W5

posicat:


blakadder:
Which wifi module is on the device now?

Issue #1394, opened 2022-04-16
[Site issue] "power monitor calibration" text always hidden on ESP32-C3 & S2


kendallgoto:
Commit [574b591](https://github.com/blakadder/templates/commit/574b591f#diff-69dca142c2e10846efa28d2c20341506e98d672c1bdead841b7c52a0d56e7ee7R233) made the "power monitor calibration" warning based on the `gpio32` variable from [_includes/pow.html](https://github.com/blakadder/templates/blob/5a62ce0208f37039c884493a9956359fed3569be/_includes/pow.html) which is only defined for `template32`. As such, power monitoring devices built on the ESP32-C3 (`templatec3`) and ESP32-S2 (`templates2`) no longer have a calibration warning. Previously, it was based on a locally defined `gpios` variable (which still exists) that aptly handles the ESP32, ESP32-C3, and ESP32-S2.

It is likely that the `pow.html` definition of `gpio32` needs to be expanded for `gpioc3` and `gpios2`, or the `templates.html` layout must be adjusted to otherwise read C3/S2 templates.

blakadder:
resolved

PR #1395, opened 2022-04-16
Add SwitchBot Plug Mini

kendallgoto:
Adds a template for [SwitchBot's Plug Mini](https://www.switch-bot.com/products/switchbot-plug-mini), the first ESP32-C3 based power monitoring plug.

As per #1394, the power monitoring calibration warning is not shown on this device, since it only has `templatec3` defined.


kendallgoto:
Also included is a Wyze / Sonoff similar OTA method to flash the device from factory. I'm not sure if there is a review process involve to officially include the repository and instructions as a part of the `flash: ...` frontmatter - please advise & I can make changes as necessary.

PR #1398, opened 2022-04-21
GPIOs off by one

grischard:
The red LED will be on permanently with the current template.

GPIO26 is the green LED, GPIO14 is the red LED.


blakadder:
thanks

PR #1402, opened 2022-04-25
Update useelink_SM-SO301AU

darkxst:
These devices have the following connections missing from the template.
Red LED on GPIO3
Button on GPIO5


blakadder:
thanks for the contribution

PR #1405, opened 2022-04-30
Update mirabella_genio_I002899

KhyberPass:
The relay GPIO for this device is 14 - not 15 that is currently documented. After programming with Tasmota I used the template that was on the site and the relay did not work. After following the tracks on the PCB I found that the relay GPIO is 14. In fact the TYWE2S does not even have a GPIO 15 pin.
I have corrected the template.
I have also created a blog post with how to open this device. You may want to link to it from the page? I was not too sure on what sort of content you wanted on these pages. If you want me to add the photos and description to the template page just let me know.
https://homeopen.wordpress.com/2022/04/27/mirabella-genio-power-plug-two-usb/


blakadder:
sure, add the link. the more info available  the better

PR #1409, opened 2022-05-03
new version of athom addressable led controller

2L3R4:
Athom seeems to have a new version of this controller since 5 Jan 2022 where the pin for ws2812 changed from gpio2 to gpio1


blakadder:
thanks

PR #1422, opened 2022-05-11
Update az-envy

frankvandenhurk:
Design flaw in the board, I think it is good to point out before people buy the board


blakadder:
thanks

PR #1432, opened 2022-05-17
New template: Adafruit QTPy ESP32-C3 dev board

fivesixzero:
Another tiny ESP32 dev board, this time an ESP32-C3 one.

Although the template worked fine, I think I might have run into one or more bugs during testing since the board seems to reset back to an unconfigured mode within a few minutes. After some looking around I don't think that's template related though.


fivesixzero:
Updated the template to remove the assignment of `GPIO9` (the board's `BOOT` button) as a `button_i`, since it seems to force a reset on this board when it's configured. Probably not a Tasmota bug, likely just the fact that it's not set up as a proper button on this board.

Setting this to `None` prevents the unexpected device resets, which is good.

fivesixzero:
Updated the template to remove the assignment of `GPIO9` (the board's `BOOT` button) as a `button_i`, since it seems to force a reset on this board when it's configured. Probably not a Tasmota bug, likely just the fact that it's not set up as a proper button on this board.

Setting this to `None` prevents the unexpected device resets, which is good.

PR #1457, opened 2022-06-04
fix MCU baudrate

fundef1:
added "SetOption97 1" to template to switch from 9600 to 115000
not sure if this is the right place but without it I couldn't get it to work.


blakadder:
that is exactly the right place, thanks

PR #1464, opened 2022-06-13
Update Youngzuth SW02-03

r2db:
This updates information regarding a specific tested configuration (transplanted ESP-12 in place of WB3S, v1.1 board). Includes information regarding a bodge wire required for full functionality with the ESP-12 module.


blakadder:
thank you

PR #1471, opened 2022-06-26
Broken URL fixes

wrt54g:
I removed some broken URLs.


wrt54g:
@blakadder Any update on this?

blakadder:
your definition of broken links is interesting. if a product is not available for purchase doesn't mean the link is broken

PR #1481, opened 2022-07-10
Update nous_A5T

clanganke:
Added comments on Tasmota update and workign around problems with the outlet buttons


blakadder:
Thanks a lot for the extra information

PR #1512, opened 2022-08-06
Microsoldering is unneeded and the gpio0 pad works

nicman23:
it just works TM

with a simple 28 awg short cable from the pad to the serial ground pad.


blakadder:
not in my experience

nicman23:
well i do not known what to tell. my soldering is baad and it still worked.

PR #1521, opened 2022-08-19
Tuya-Convert in 2022 for Gosund EP2

EpicLPer:
Edited the Gosund EP2 description and added notes regarding a working "workaround" for Tuya-Convert which involves sending a mail to Gosund and requesting a firmware upgrade manually.


EpicLPer:
Finally requested a pull cause apparently I completely forgot about this a month ago, whoops...

PR #1532, opened 2022-08-29
Update BSD34

iPhilBln:
Added the template config for power meshurement with BL0937


blakadder:
where did you buy the version with power monitoring? is it still marked BSD34?

PR #1546, opened 2022-09-15
Fixed a few metadata anomalies

sfromis:
Noticed a few deviations from expected template layout, I assume that my changes are uncontroversial


sfromis:
@blakadder  I see that you removed the initial `---` from `smartmi_ZNNF072M` again. To me, adding it looked like an unambiguous fix to be done, as all the other templates have `---` before and after the metadata block with yaml-like key/values. Am I missing something there?

PR #1558, opened 2022-09-27
Add failed eBay convert for Gosund UP111

mstratford:


blakadder:
thanks for sharing the info

PR #1559, opened 2022-09-28
Update ce_smart_home_LTS-6A-W5

dac2020:
I didn't want to replace the previous information but add this.

To have your plugs in order, starting left to right (cable end being outlet 1), use this template.
{"NAME":"CE Power Strip","GPIO":[288,0,0,0,227,228,0,0,225,226,224,0,32,0],"FLAG":0,"BASE":18}


blakadder:
overwriting is ok

PR #1561, opened 2022-09-29
Search fixes

AndreKR:


blakadder:
thanks a lot, works much better than before!

PR #1570, opened 2022-10-07
Add amazon purchase in October

VizuaaLOG:
Adds a recent successful flash with an Amazon purchase


blakadder:
Thanks for the information

PR #1571, opened 2022-10-12
Add Shopping Website Code in html and css

Chirag0407:
Add Shopping Website Code in html and css


blakadder:
nice try, i respect the hustle but wth

PR #1575, opened 2022-10-16
Wipro RGB Tube Light - Inverted Color Temp Fix

nisgoswami:


nisgoswami:
This is a fix for the Wipro RGB Tube Light , changing the template fixes the inverted color template

PR #1581, opened 2022-10-19
Update template

arendst:


sfromis (changes requested):
When updating the template to new component numbers, the field name should be "template9" instead of just "template" :-)

PR #1584, opened 2022-10-21
Update nedis_WIFIDS10WT

domerich:
additional information, I struggled without it.


blakadder:
That is basic usage information covered in official Tasmota documentation.

PR #1586, opened 2022-10-21
Update with new date and more accurate serial flashing info.

blacktirion:


blacktirion (commented):
Clarifications

blakadder:
thanks

PR #1591, opened 2022-10-27
Add alternative template from S26 to S26R2

niwla23:
The information about the alternative template was only present in the S26 page but not in the S26R2 page. This PR adds it to S26R2 too. I have tested it and it works also on the S26R2


blakadder:
Thanks

PR #1602, opened 2022-11-12
Update lenovo_ZA7F0000WW with power monitoring

kohrar:
Added missing pins for HLWBL, BL0937, and blue LED for the Lenovo ZA7F0000WW smart plug.

I managed to get Tasmota without needing to desolder the module. The pads next to the TX/RX pins can be accessed by scraping the solder mask off. I've documented this on my personal wiki at https://leo.leung.xyz/wiki/Lenovo_SE-341AC#Wiring in case you want to take a look.

I'm not 100% sure how to get the blue LED to turn on when the swith is off and also have the red LED turn on when the switch is on. As a result, I've assigned the blue LED as Led_i 3 for now.


blakadder:
thanks for the extra info

PR #1606, opened 2022-11-17
Create oakter_oak_plug_mini

RajeGanesh:
F793C2C3-402A-4B05-857D-64F38FD202A9
Added images inline with orientation description.


blakadder:
empty commit?

PR #1610, opened 2022-11-21
Update and rename shelly_DUO_RGBW to shelly_DUO_RGBW_E27

gzone156:
Added several resellers links
Updated the manufacturer link to a more specific page from the manufacturer's KB Better explanation of WhiteBlend and VirtualCT modes to be enabled for a better experience


blakadder:
Same as the other PR and renaming the file breaks the link

PR #1614, opened 2022-11-22
Update arilux_LC11 with new link on guide

cj-wong:
The previous link no longer points to the right location. I have replaced it with the new (canonical) link.


blakadder:
thanks

PR #1617, opened 2022-11-25
Modify the correct indicator light

tarontop:
Because one indicator light needs to indicate relay status and Wi-Fi status


blakadder:
the current template is the default template that comes from athom, you're free to modify it for yourself but that doesn't make it the default one

PR #1620, opened 2022-11-27
added Wemos D1 Mini ESP32; fine-tuned Wemos Lolin32 Lite (ESP32)

heikokue:
I hope, all is correct for a new module


blakadder:
looks good, thanks

PR #1629, opened 2022-12-06
Update Moes_WHT-HY609

andreas-nn:
new found dplds 102,119-122
some examples


luusl:
Hi, could you please provide your user_config_override.h when compiling the tasmota binary. I've flashed the original tasmota.bin, but it's pretty useless. After reading https://tasmota.github.io/docs/Thermostat/ I've found I need to build my own binary to use this. Hopefully you can provide a litte bit more info to get this working. Thank you.

PR #1650, opened 2022-12-23
fix: typo in Lerlink URL

Eelviny:
Accidentally left a `}` in #1648


Eelviny:
also added fix to make sure all Lerlink entries are consistent, some were marked as only EU when they're also UK

blakadder:
great! thanks

PR #1654, opened 2023-01-05
Update martin-jerry-3-way-ST01

MaceMoneta:
1. Also applies to MJ-ST02 (Tasmota pre-installed)
2. Module must be configured to: Tuya MCU (54)
3. Tuya mapping required in console: TuyaMCU 11,1


blakadder:
This commit makes no sense because the original template doesn't have an MCU. MJ-ST01 and MJ-ST02 are two different devices and cannot share a template

MaceMoneta:
I wasn't aware they were that different in the hardware. Closing.

PR #1682, opened 2023-01-30
Update petoneer_TY011

ov1d1u:
I got a Petoneer TY011 but the PCB looks a bit different than the other in the pictures, and the steps described here don't work. I added a mention about this so other would be warned to also check for PCB similarities.


blakadder:
do you have images of the product label and the PCB?

ov1d1u:
@blakadder  Yes:

PXL_20230130_090214749

PR #1684, opened 2023-02-07
Fix aliexpress link for athom rgbw controller

apbarratt:
The original link was pointing at athom's rgb controller, rather than the rgbw one :)


blakadder:
and the updated link links to an addressable LED controller which isn't an RGBW controller

apbarratt:
Whoops, sorry, I'd intended this link, I had many tabs open at the time and mixed up WLED with RGBW in my head 😅
https://www.aliexpress.com/item/1005003796227040.html

apbarratt:
That's it fixed now :)

apbarratt:
Oh, or not, looks like PRs don't autoupdate with new commits like on gitlab.  Ah well, it's not important anyway.

PR #1686, opened 2023-02-13
Added instructions for reseting the martin jerry ST01

michaelkamprath:
The Martin Jerry ST01 3-way switch pre-flashed with Tasmota actually has a Tayu MCU that controls the relay. Updating Tasmota or otherwise resetting it on the device will remove the configuration that allows the ESP8266 to communicate with the Tayu MCU. This update adds instructions for how to restore a working configuration.

### DIFF
--- a/_templates/martin_jerry_ST01
+++ b/_templates/martin_jerry_ST01
@@ -13,3 +13,15 @@ type: Switch
 standard: us
 ---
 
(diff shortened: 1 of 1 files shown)


michaelkamprath:
I am realizing my Tasmota template needs more work. Contacting Martin Jerry for help. Will reopen this PR when it is ready to go.

PR #1687, opened 2023-02-13
Update lilygo_T-Camera

Tirguy:
Correction : old template did not work, this one can now send pictures; put the model in the name too ("A5")


blakadder:
Old template didn't work because you have a different device so you should create a new page for yours

PR #1690, opened 2023-02-17
Update tomzn_TOB8-63WiFi

kodiakastu:
Added chip field and a little description on how to open the device and have access to the mcu.


blakadder:
Thanks

PR #1692, opened 2023-02-24
Update ewelink_ZB-GW03

xsp1989:
Update download link


blakadder:
not an official source and not automated

PR #1702, opened 2023-03-09
Added L8 power meter gpio settings and updated description

Silverstar:
I just obtained a L8-HS EU and wanted to get power metering working.
To check which gpio it is connected to and which chip is used, I measured all unknown (to me) pins which are exposed through the socket on the back.

"CF": IO35 to HLW8012 CF
"HUMI": SENSOR-VN, not connected on the mains pcb
"BP": via R33, Q8, R32 (and R31 to GND) to IO15, not connected on the mains pcb
Either BP or HUMI is in fact connected to a unpopulated spot "P1" next to the connectors to the logic pcb, by silk size and shape (circle) possibly a capacitor, maybe necessary in the dimmer or other variant. The silk shows that the pcb has many different equipping options.

Voltage and Current metering are not available because the CF1 and SEL pins are not connected anywhere on the mains pcb.

With a power meter plug and a 25W halogen bulb I obtained readings of 27.xx W on the power meter and with PowerSet I received 27 W on the web interface, sometimes jerking to 14 or 54 W for less than a second, but most of the time shows 27 W correctly.

### DIFF
--- a/_templates/lanbon_L8-HS
+++ b/_templates/lanbon_L8-HS
@@ -6,7 +6,7 @@ category: switch
 type: Display Switch
 standard: [eu, us]
 image: /assets/device_images/lanbon_L8-HS.webp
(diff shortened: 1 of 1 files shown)


blakadder:
Thanks

PR #1726, opened 2023-04-19
Update OV5640_V1_2

sfromis:
Clearly should be "tasmota32".....


smartboart:
Hello,
is this template working? For me it is not...

sfromis:
IDK, but my starting expectation would not include Tasmota having support for OV5640 camera sensor

PR #1731, opened 2023-04-28
M5Stack AtomS3 Lite ESP32-S3 template

lboue:
Add M5Stack AtomS3 Lite ESP32-S3 template


lboue:
@blakadder Would it be possible to validate the addition of this new device?

blakadder:
It would

PR #1733, opened 2023-04-30
Update dingtian_DT-R008

jeroenst:
Added documentation for hardware version 3.6.10 and higher 
Applied better formatting of document


blakadder:
thanks

PR #1747, opened 2023-05-29
Add template for Wemos LOLIN C3 PICO V1.0.0

lboue:
Add template for Wemos LOLIN C3 PICO V1.0.0
https://www.wemos.cc/en/latest/c3/c3_pico.html


lboue:
Thanks

PR #1749, opened 2023-06-11
Update duux_DXCF10

barrymossel:
Correct template, as in: https://github.com/arendst/Tasmota/issues/18854


blakadder:
you've removed the entire template with this

PR #1758, opened 2023-07-09
Add Shelly Plus Plug UK

imduffy15:
Template unknown

Pending https://github.com/tasmota/mgos32-to-tasmota32/pull/3 for the firmware build.


blakadder:
so what's the point in adding it?

imduffy15:
To let folks know that it is flashable.

blakadder:
if you flashed it you should have a template

PR #1761, opened 2023-07-18
Update flash sizes listed for ESP8685-WROOM-06

hendrikvh:
The flash size of the H2 and H4 was incorrect. The ESP8685-WROOM-06-H2 has 2MB flash and the H4 has 4MB. See the datasheet page 9 here: https://www.espressif.com/sites/default/files/documentation/esp8685_datasheet_en.pdf

comparison


blakadder:
thanks, had to update every module so this is covered in another commit

PR #1771, opened 2023-08-26
Update ucomen_PA-GEH-01SWP2

marceld83:
A've added a short description on dissasembly of the module. If you could help me, I have made pictures to help future dissasembly. However, I'm do not have any exprecience with GitHub as to include pictures in the discription.


blakadder:
you can add images to the comments here and I will take care of the rest

PR #1775, opened 2023-09-07
Update ZY-M100 rules

pedronsg:
Added illuminance rule


blakadder:
Iluminance is already configured through TuyaMCU command

PR #1786, opened 2023-10-07
Update gosund_WP6 (bl0937)

MissingTwins:
The original template was kind off to bl0937(Meter Chip)


blakadder:
Gosund WP6 does not support power monitoring

MissingTwins:
@blakadder 

But it does, at least my variant does. It is equipped with BL0937.
I bought WP6 primarily because it was capable of power monitoring and also came with an ESP8266 inside.

image
image
image
image

blakadder:
It might be called WP6 but its a new device since the existing wp6 is a US plug. You'll need to submit it as a New device

PR #1788, opened 2023-10-16
Shelly Plus 1 UL is dual core

marcusb:
Some Shelly Plus 1 devices (in my case the UL version) are dual-core and can be flashed with the regular tasmota32 binary.


blakadder:
thanks

PR #1789, opened 2023-10-16
Update wipro_NS7001

Amit-Shan:
I overlooked the instructions on the template GitHub docs asking me to edit existing device templates if I found an error. So, you will probably find a google form from me in the add new device section for the same device. Please disregard that form. My sincerest apologies for creating two different entries for the same device. 

Please find the updated templates and device information in the edit.

Edit Information:
1. Adjusted Bulb type to RGBW from RGB
2. Adjusted amazon link to point to the correct device. It previously pointed to what appears to be a newer bulb from the same brand.
3. Adjusted template type to the latest version of Tasmota.
4. Replaced template with the correct one. The new one removes PWM5 from the GPIO settings as the bulb does not support White Color Temperature. Enabling PWM5 causes the bulb to behave unnaturally. 
5. Added flash method as Tuya-Convert. It's what I used a few days ago. It does work for this model and revision. I do not think the manufacturers has updated the firmware ever, so this particular model should still work with Tuya-Convert. That being said, there may be other factors at play like Linux and Python versions and device revisions I am unaware of.

### DIFF
--- a/_templates/wipro_NS7001
+++ b/_templates/wipro_NS7001
@@ -1,11 +1,13 @@
 ---
 date_added: 2019-11-23
 title: Wipro Garnet NS7001 480lm
-type: RGB
+model: NS7001
+type: RGBW
 category: bulb
 standard: b22
-link: https://www.amazon.in/Enabled-12-Watt-Million-Compatible-Assistant/dp/B07S5TH6BN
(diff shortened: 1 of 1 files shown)


blakadder:
added it from the form submission

Amit-Shan:
Sounds good. Apologies for the duplicate submissions. Will use the proper channels if I ever come across anything else. Thank you.

PR #1790, opened 2023-10-17
Add flashing instructions

andySFRD:
- Add flashing instructions
- Fix manufacturer URL
- Remove duplicate setOption commands


blakadder:
thanks

PR #1826, opened 2023-12-19
Update sonoff_T5-4C-86

babybambula:
Changed configuration from 1 gang to 4 gang by adding the missing relays to the respective GPIOs.
Tested it on my device, works fine. 

Added a comment about the last switch in the GUI.


blakadder:
thanks

PR #1827, opened 2023-12-22
Update wion_50053 link

cscott:
The article linked is no longer at the posted URL, but archive.org has a copy.  Seems to still apply to this hardware.


blakadder:
thanks

PR #1831, opened 2023-12-29
add sunton 7 inch

markusdd:
add sunton 7 inch display
change template name to make difference clear in sunton 5 inch

relates to this issue in Tasmota: https://github.com/arendst/Tasmota/issues/20341#issuecomment-1872259273

Tasmota autoconf-PR to be consistent with this PR: https://github.com/tasmota/autoconf/pull/39

The boards are identical except for the display driver IC, which changes the pinout.
So the Tasmota template is the same, but the display.ini differs and hence does the autoconf package.


markusdd (commented):
what I forgot: I recommend to re-use the 5 inch image.

From these pictures it is impossible to tell if it is the 5 or 7 inch so we might as well...

PR #1832, opened 2023-12-30
Update kogan-KASMCDSKTLA

JAMESBOWLER:
Added OpenHAB configuration 

Mainly so if it all crashes I can find info here


JAMESBOWLER:
commit request

PR #1843, opened 2024-01-26
Update shelly_plus_1 with DS18x20

Kolia56:
Add a template to support newly added DS18x20 support


blakadder:
added as a separate device

PR #1854, opened 2024-03-05
Update ghome_EP2-A

rovo89:
Switch Led_i and LedLinki. After the change, the power status is shown with the green LED and the WiFi status with the blue LED.

### DIFF
--- a/_templates/ghome_EP2-A
+++ b/_templates/ghome_EP2-A
@@ -3,7 +3,7 @@ date_added: 2023-12-09
 title: GHome EP2-A
 model: EP2-A
 image: /assets/device_images/ghome_EP2-A.webp
-template9: '{"NAME":"GHome EP2-A","GPIO":[320,0,576,0,2656,2720,0,0,2624,32,0,224,0,0],"FLAG":0,"BASE":45}' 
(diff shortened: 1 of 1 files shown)


barbudor:
As far as I understand, this change is not functional but a personal preference

I suggest that you modify your PR to add this template in the text and explaining the difference between both so user can choose which one they prefer

rovo89:
I can do that if you prefer, although I always associate the colors this way.
Do you have an example how this can be achieved, especially with the explanatory text?

barbudor:
@rovo89 you can add text below
Skip one line and you can add mark down formated text

See that one
https://github.com/blakadder/templates/blob/master/_templates/rocket_R4A4

rovo89:
I took some inspiration here: https://github.com/blakadder/templates/blob/master/_templates/blitzwolf_SHP6

rodajc:
With the modification as proposed by @rovo89 the green LED indicates the power state which is definitely more consistent with other devices and with the original firmware.

When I submitted the first version of the template for this device I copied it from the old Nous A1 as recommended by... @barbudor.

I didn't pay much attention to the LEDs as I didn't know what to expect. I was happy it was switching on and off and submitted the template after changing only the NAME.

Now @rovo89 proposes an enhancement I think we should publish as the template to use with this device.

PR #1862, opened 2024-03-29
Seeed Studio XIAO ESP32C6

lboue:
Add support for Seeed Studio XIAO ESP32C6 board.


lboue:
Hello @blakadder,

Is anything missing before validating?

Regards

PR #1872, opened 2024-05-03
Update ESP-C3-12F pictures

lightmaster:
Updated ESP-C3-12F picture with better photo of one.
Updated ESP-C3-12F_pinout to correct several incorrectly labeled GPIO pins and to show several alternative pin labels.


blakadder:
Those are not the same modules

PR #1912, opened 2024-12-24
Update ESP32_Relay_X8

DraconPern:
Reversed the order of the numbering of the relay to match what's on the silkscreen.

also, the ebay link is dead, however a quick search shows a lot of resellers.  Not sure which one is appropriate to link so I didn't change it.


DraconPern:
After making a project using this board, I used the original template.

Claus-E:
@DraconPern was right. The Relay numbering is wrong in the template form 2022. And in Addition we can add the LinkLED On GPIO23.

PR #1916, opened 2025-01-14
Add Mursche Infrared Heater

DaKarlii:
Added new Device Mursche Infrared Heater


DaKarlii:
@blakadder any comments / suggestions on that PR?

PR #1931, opened 2025-03-19
Update deta_6912HA

DrWorm-AU:
updated template


DrWorm-AU:
updating the template to current standard

PR #1941, opened 2025-05-31
Update athom_EM2

mvdw:
Formatting only - this makes the last part (hopefully) readable on https://templates.blakadder.com/athom_EM2.html


mvdw:
Yes! It  worked.
Thanks.

PR #1952, opened 2025-07-20
Update WD-01ADE

yvesdm3000:
Added info on the newer WD-01ADE using an ESP8685-WROOM-06

### DIFF
--- a/_templates/WD-01ADE
+++ b/_templates/WD-01ADE
@@ -38,3 +38,12 @@ For use with Home Assistant its advised to set friendly names for the relays:
 ```console
 Backlog FriendlyName1 Pump A; FriendlyName2 Pump B
 ```
+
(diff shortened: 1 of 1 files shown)


rannooo:
@yvesdm3000 can you explain the added comment?

> The board has been redesigned and is no longer compatible with a PSF-B04. Luckily an ESP8685-WROOM-06 still fits altough 1 pin is mapped to a “NC” pin, so I had to add a wire to another pin. This pin is used for the blue “WLAN” led.

Does this mean that the newer CBU chip needs to be changed by soldering or just that the ESP8685-WROOM-06 can be used for flashing?

blakadder:
Yes

rannooo:
😂 yes what - new chip re-soldering needed or not needed?

PR #1959, opened 2025-09-10
Update sunton_ESP32-2432S028

jsiddall:
Updated template and display.ini for current Tasmota releases (starting about release 14) that include uTouch


pandlab:
Tested, works, thanks

PR #1993, opened 2026-04-17
Update agl_1106063 with device notes and instructions

tiagodamian:
Added notes about device compatibility and assembly instructions.


tiagodamian:
Couldnt add image to comiit.

